PR #24, opened 2019-03-20
[WIP] cmd/govim: add AST-based highlighting

myitcv:
Demo: https://youtu.be/OGtby_bjNUY

TODO:

* work out what bug is causing some `matchaddpos` calls to "stall"
* we should throttle buffer updates
* make it work for raw string literals. Currently does not work because the highlight is line-oriented, and raw string literals can span multiple lines
* batch calls to `matchaddpos` and `deletematch`:
  * group `matchaddpos` calls into line-based updates
  * make the calls to `matchaddpos` silent so that Vim does not redraw after each one
  * explicitly redraw once done
* only re-highlight when the view window changes
* raise a Vim feature request for an event to be fired when the viewport changes, where "viewport" is defined (for now) as `getwininfo()` minus the `"variables"` key


myitcv:
Closing for now. We can pick this up again later if necessary.

PR #34, opened 2019-03-27
cmd/govim: call DidOpen and DidChange for buffers

myitcv:
Also fixed the UTF-16 position information


myitcv:
Travis seems to be having a bad day. This all passes locally.

PR #35, opened 2019-03-28
govim: refactor to define concept of a Plugin

myitcv:
Also do some renames in testdriver to remove potential for confusion in
what a Driver is.


myitcv:
Screen Shot 2019-03-28 at 13 02 24

PR #36, opened 2019-03-28
govim: add buffer number to the viewport and fix JSON parsing

myitcv:


myitcv:
Travis still slow. Passes locally.

PR #37, opened 2019-03-28
cmd/govim: implement BufWritePre hook to format source

myitcv:


myitcv:
Docker still being slow

PR #38, opened 2019-03-28
cmd/govim: revendor x/tools/internal

myitcv:


myitcv:
Travis still being slow

PR #85, opened 2019-04-06
cmd/govim: add a test for when a new file is created from a new buffer

myitcv:


myitcv:
Blocked on https://github.com/golang/go/issues/31522 (per https://github.com/myitcv/govim/issues/86)

PR #95, opened 2019-04-08
[WIP] cmd/govim: test that verifies completion works with unpersisted imports

myitcv:
See https://github.com/myitcv/govim/issues/96


myitcv:
Blocked on golang/go#31338

myitcv:
Can potentially be worked around with a solution to https://github.com/golang/go/issues/31796

PR #102, opened 2019-04-08
ci: add gvim to CI matrix

myitcv:


myitcv:
With thanks to @chrisbra

PR #103, opened 2019-04-09
deps: use latest tools

myitcv:
Also add a test to verify a file with syntax errors saves ok.


myitcv:
Currently blocked on https://github.com/golang/go/issues/31359

Issue #121, opened 2019-04-15
doesn't work on neovim?


gracenoah:
I see the following on starteup in neovim:
```
Error detected while processing /home/gracenoah/.config/nvim/plugged/govim/plugin/govim.vim:
line   13:
E518: Unknown option: ttymouse=sgr
line   14:
E518: Unknown option: balloondelay=250
line   15:
E518: Unknown option: ballooneval
line   16:
E518: Unknown option: balloonevalterm
line  286:
E117: Unknown function: job_start
E15: Invalid expression: job_start(start, opts)
Press ENTER or type command to continue
```

```
nvim --version
NVIM v0.4.0-150-g279043d62
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: /usr/bin/gcc-5 -Wshadow -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -DMIN_LOG_LEVEL=3 -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector-strong -fdiagnostics-color=auto -Wno-array-bounds -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/home/travis/build/neovim/bot-ci/build/neovim/build/config -I/home/travis/build/neovim/bot-ci/build/neovim/src -I/home/travis/build/neovim/bot-ci/build/neovim/.deps/usr/include -I/usr/include -I/home/travis/build/neovim/bot-ci/build/neovim/build/src/nvim/auto -I/home/travis/build/neovim/bot-ci/build/neovim/build/include
Compiled by travis@travis-job-5936a537-306c-42ad-aea4-b95ece112e81

Features: +acl +iconv +jemalloc +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/share/nvim"

Run :checkhealth for more info
```

myitcv:
Hi @gracenoah - correct, Neovim is not (currently) supported, per https://github.com/myitcv/govim/wiki/FAQ#what-versions-of-vim-and-go-are-supported

But I think it makes sense to have an issue open where this can be discussed.

I personally do not have plans to add Neovim support, but I certainly wouldn't be averse to someone investigating this.

[...]

PR #186, opened 2019-04-28
[WIP] [experimental] Add Neovim support

milesbxf:
This is an experiment to explore how we could add support for Neovim. As suggested in #121, this adds a new implementation of [`govim.Govim`](https://godoc.org/github.com/myitcv/govim#Govim) that supports the Neovim RPC API, along with some Vimscript and testdriver modification to get the govim tests (in `testdata`) working.

I haven't tried this at all with the `cmd/govim` plugin itself, but I suspect it won't quite fully work yet - if we're happy with this approach, get it cleaned up and merged in then I guess that would be the next step.

**Govim interface differences**

The RPC event models of Vim8/Neovim are actually relatively similar, with async bidirectional communication and event handlers. Neovim exposes a really powerful RPC interface itself which allows external programs/plugins to drive pretty much anything, including directly calling Vimscript functions - more on this below.

Fortunately my work was made significantly easier by discovering the [neovim/go-client](https://github.com/neovim/go-client) project, which provides this interface as a Go library, abstracting away all the nasty event handling and msgpack stuff. It has its own [`plugin.Plugin`](https://godoc.org/github.com/neovim/go-client/nvim/plugin#Plugin) interface, which I guess has a similar aim to `myitcv/govim` to expose a high-level API for plugin authors, but I was able to use the more low level [`nvim.Nvim`](https://godoc.org/github.com/neovim/go-client/nvim#Nvim) API to adapt it for our purposes. I've copied/cargo culted quite a bit of logic from the original `goVimImpl` methods, but we can deduplicate and clean this up later.

**Vimscript differences**

The primary difference is with the RPC APIs - `jobstart` instead of `job_start` (with different `opts`), `rpcrequest` instead of `ch_evalexpr`. There's a difference in the structure of the actual request objects and responses that I haven't quite pinned down (though what we have here works well enough) - other than that, the msgpack/JSON protocols don't seem to make a difference.

One significant improvement is that we're able to directly call out to Vimscript functions from the plugin, so I've defined a few public API wrappers (e.g. `govim#defineCommand`).

**Testdriver differences**

To my great surprise, I hardly had to change the test infrastructure at all. Neovim expects its `.vimrc` and plugins in totally different locations, so we switch on that (plus the executable name of course).

## TODO

- [ ] Get a Neovim image on CI so we have automated testing against this -> #211 
- [ ] implement viewport functions - it will just panic at the moment (could we write tests for these?)
- [ ] investigate flaky `testscripts/unscheduled.txt` test (seems to fail around 1 in 6 times, some kind of race condition writing out)


myitcv:
This is great, thanks @milesbxf. Let's continue discussion in #121 for now to see whether we get a strong steer from others on direction, etc. That said, having this PR, even if it subsequently changes shape, as a basis for discussion is very helpful in making certain points concrete.

myitcv:
@milesbxf - in the meantime, I think it probably makes sense to:

* create another PR in which we make the required changes to add Neovim to the Docker image so that...
* ... in this PR we can also add Neovim to the CI matrix

Thoughts?

milesbxf:
@myitcv 👉 #211 for CI images

milesbxf:
CI loads fine (hooray!), but the Neovim tests are timing out. I haven't run them locally for a while, so potentially something has changed on master, will take a look.

plugin/govim.vim
```diff
@@ -8,14 +8,26 @@ let s:tmpdir = $TMPDIR
...
+    let s:ch_logfile = trim(s:ch_logfile)
+    call ch_logfile(s:ch_logfile, "a")
+    echom "Vim channel logfile: ".s:ch_logfile
+    call feedkeys(" ") " to prevent press ENTER to continue
+
+    "Balloon popups aren't supported by Neovim yet
+    set balloondelay=250
+    set ballooneval
+    set balloonevalterm
+    set updatetime=500
```
milesbxf: Move this line out

neogovim.go
```diff
@@ -0,0 +1,487 @@
...
+func (n *NeoGovim) UnsubOnViewportChange(sub *OnViewportChangeSub) {
+	panic("UnsubOnViewportChange() not implemented yet")
+}
+
+func (n *NeoGovim) ToggleOnViewportChange() {
+	panic("ToggleOnViewportChange() not implemented yet")
+}
+
+// Viewport returns the active Vim viewport
+func (n *NeoGovim) Viewport() (Viewport, error) {
```
milesbxf: Focus on implementing this, don't worry too much about other viewport methods

myitcv:
Apologies, this PR got automatically closed as a result of the changes for #910 - we should have changed the base of open PRs before deleting the `master` branch. Please can you change the base of this PR from `master` to `main`?

PR #189, opened 2019-04-30
cmd/govim: temporary implementation of file watching

myitcv:
For now use textDocument/didOpen and textDocument/didChange to deal with
generated files changing, whilst we await a proper implementation in
gopls via golang/go#31553.


myitcv:
Currently blocked on https://github.com/golang/go/issues/31759; although we could rework the tests to workaround the test failure. It is a legitimate test failure however.

PR #206, opened 2019-05-03
deps: golang.org/x/tools v0.0.0-20190503030157-5cec639030af

myitcv:
* internal/span: add a filename only print for spans
* internal/lsp: support builtin types without hardcoding
* internal/lsp: first pass at some hover tests
* internal/lsp: generate RPC interface from Typescript source
* internal/lsp: change some comments and variable names in completion code
* internal/lsp: clean up formatting header handling
* internal/lsp: handle nil pointer in (*Package).GetActionGraph


myitcv:
Build failures are either flakes because of https://github.com/golang/go/issues/31822 or timeouts between Travis and GitHub.

PR #213, opened 2019-05-04
ci: switch to use govim DockerHub org

myitcv:


myitcv:
Some timeouts from Travis to GitHub again... but otherwise fine.

PR #215, opened 2019-05-06
cmd/govim: investigate complete_watched test flake

myitcv:


myitcv:
Other than timeout flakes (https://github.com/myitcv/govim/issues/216)

PR #236, opened 2019-05-19
testdriver: improve debug logging

milesbxf:
This makes it easier to debug issues with the govim tests, by allowing
users to log more information with a general -debugLog flag. This
enables Vim debug logging, and sends Vim startup output to the normal
Govim log. If Vim fails to start, it prints out the contents of the log,
which should hopefully contain some useful information. It also prints a
full command to run Vim with the temporary home directory.

Since the call to NewTestDriver was getting a bit unwieldy with over 10
args, I refactored that out into a Config struct.


myitcv:
One small request: with commits, PRs and issues, please can you follow the following approach?

https://golang.org/doc/contribute.html#commit_messages

Specifically:

> The first line of the change description is conventionally a short one-line summary of the change, prefixed by the primary affected package.

For example, here are the commits in `master`:

https://github.com/myitcv/govim/commits/master

It just means it's then trivial to create a relatively brief log of what's happened between versions:

```
$ git log --pretty=format:'* %s'
* cmd/govim: allow undo before format-on-save (#252)
* cmd/govim: change default of GOVIM_DISABLE_INCREMENTALSYNC to true (#250)
* cmd/govim: add support for delta-based buffer updates (#239)
* scripts: tidy and fully add Neovim image and flavor (#249)
* github: add funding details as a test (#246)
* ci: make GH_USER and GH_TOKEN optional (#245)
...
```

testdriver/testdriver.go
```diff
@@ -110,13 +133,47 @@ func NewTestDriver(name string, govimPath, testHomePath, testPluginPath string,
...
 	return res, nil
 }
 
+func (d *TestDriver) Logf(format string, a ...interface{}) {
+	fmt.Fprintf(d.Log, format+"\n", a...)
+}
+func (d *TestDriver) LogStripANSI(r io.Reader) {
+	buf := bufio.NewReader(r)
+	for {
+		l, err := buf.ReadString('\n')
```
myitcv: Beware! In https://github.com/myitcv/govim/pull/299 I effectively fell foul of this default `bufio` buffer overflowing because of long lines. What do you actually use this output for?

testdriver/testdriver.go
```diff
@@ -110,13 +133,47 @@ func NewTestDriver(name string, govimPath, testHomePath, testPluginPath string,
...
+		}
+
+		fmt.Printf("Test command:\n==========================\npushd %s && %s %s && popd\n==========================\n", c.Env.WorkDir, envlist, strings.Join(res.cmd.Args, " "))
+	}
 
 	return res, nil
 }
 
+func (d *TestDriver) Logf(format string, a ...interface{}) {
+	fmt.Fprintf(d.Log, format+"\n", a...)
```
myitcv: Again, only output if debugging enabled?

govim_test.go
```diff
@@ -52,37 +52,70 @@ func TestScripts(t *testing.T) {
 					"HOME="+home,
 				)
 				testPluginPath := filepath.Join(e.WorkDir, "home", ".vim", "pack", "plugins", "start", "govim")
-				d := newTestPlugin(plugin.NewDriver(""))
-				td, err := testdriver.NewTestDriver(filepath.Base(e.WorkDir), ".", home, testPluginPath, e, d)
-				if err != nil {
-					t.Fatalf("failed to create new driver: %v", err)
-				}
+				fmt.Printf("Vim home path is at %s\n", home)
```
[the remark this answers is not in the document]
milesbxf: Yes, very good point - I'll go through and vet the `printf` calls

testdriver/testdriver.go
milesbxf: This should only be going to `d.Log`, which by default only logs to an error file

testdriver/testdriver.go
milesbxf: Ah, good catch, though I haven't come across this in testing.

I've found this very useful when debugging Vim startup errors

govim_test.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 var (
-	fLogGovim = flag.Bool("govimLog", false, "whether to log govim activity")
+	fDebugLog = flag.Bool("debugLog", false, "whether to log debugging info from vim, govim and the test shim")
```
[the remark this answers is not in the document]
myitcv: This looks to be resolved.

PR #253, opened 2019-05-29
cmd/govim: fix and re-enable deltas; apply format updates in batches

myitcv:
This way, we always make changes to buffers in Vim/VimScript. This
avoids the scenario described in
https://github.com/vim/vim/issues/4438#issuecomment-496448888.


myitcv:
This is currently blocked on https://github.com/vim/vim/issues/4438#issuecomment-496887725

PR #291, opened 2019-06-12
cmd/govim: add GOVIMReferences command for gopls References method

myitcv:


myitcv:
Blocked on https://github.com/golang/go/issues/32572

PR #294, opened 2019-06-12
cmd/govim: only update quickfix when user is not busy

myitcv:


myitcv:
FYI @leitzler - this should make the quickfix updates less distracting.

PR #326, opened 2019-06-29
cmd/govim: add test for quickfix errors in dependencies

myitcv:


_scripts/runDockerRun.sh
```diff
@@ -8,8 +8,9 @@ cd "${BASH_SOURCE%/*}/../"
 
 # The ARTEFACTS variable set by .travis.yml cannot expand
 # variables so we do that here
-ARTEFACTS=$(echo $ARTEFACTS)
+ARTEFACTS=$(echo ${ARTEFACTS:-})
 
+artefacts=""
```
leitzler: leftover?

go.sum
```diff
@@ -43,6 +43,7 @@ github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81P
 golang.org/x/crypto v0.0.0-20181106171534-e4dc69e5b2fd/go.mod h1:6SG95UA2DQfeDnfUPMdvaQW0Q7yPrPDi9nlGo2tz2b4=
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
 golang.org/x/crypto v0.0.0-20190510104115-cbcb75029529/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
+golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550 h1:ObdrDkeb4kJdCP557AjRjq69pTHfNouLtWZG7j9rPN8=
```
leitzler: I'm a bit curious why this was added now and not earlier since nothing else changed, and we do `go mod tidy` as a part of CI?

_scripts/runDockerRun.sh
myitcv: No, this is necessary to avoid the "use before initialised" error that otherwise would occur on line 26 when used outside of CI.

go.sum
myitcv: Hmm, I hadn't spotted this. Will investigate.

myitcv: I'm a bit perplexed TBH. I'll merge #691 and create a Go issue of the back of your observation.

myitcv: Raised https://github.com/golang/go/issues/36667

PR #338, opened 2019-07-07
[WIP] govim: refactor transport/protocol into new package

milesbxf:
This adds a new `github.com/myitcv/govim/internal/transport` package with a `Transport` interface. The aim here is to make `govim.Govim` generic enough that it can support both Vim and Neovim (#186) with as few code changes as possible, whilst handling the differences in protocol (JSON vs msgpack) as much as possible in the transport layer.

This should make it much easier to add support for Neovim, and minimising the amount of separate code we have to maintain for both should make things easier in the future.

This is still very much WIP, there's still quite a few things left to refactor/clean up:

- [x] The govim run loop still has the concept of a callback ID (to respond to Vim) - this should be hidden away in the transport layer
- [ ] Error handling. The original `Govim` interface has a nice distinction between user-level errors (e.g. "vim function called with wrong arguments) vs protocol-level errors (e.g. "RPC connection closed). We should find a way of keeping this
- [x] `transport.Transport` requires an event queue object passed in to queue things. It only needs this to enqueue responses to scheduled callbacks - perhaps we can just pass in a "schedule" function
- [ ] The `transport.Transport` interface should be wire-format agnostic - callees should pass in and receive Go objects rather than `json.RawMessage`

Early feedback welcomed.

### DIFF
--- a/internal/transport/types.go
+++ b/internal/transport/types.go
@@ -20,7 +20,6 @@ type Transport interface {
 	Initialized() chan struct{}
 	IsShutdown() chan struct{}
 
-	Read() (responseFunc func(p2 interface{}, ps ...interface{}) error, messageType string, args []json.RawMessage, err error)
 	Send(callback Callback, callbackType string, params ...interface{}) error
 	SendAndReceive(messageType string, args ...interface{}) (json.RawMessage, error)
 	SendAndReceiveAsync(messageType string, args ...interface{}) (ScheduledCallback, error)
--- a/internal/transport/types.go
+++ b/internal/transport/types.go
@@ -1,0 +1,28 @@
+package transport
+
+import (
+	"encoding/json"
+	"errors"
+)
+
+var (
+	ErrShuttingDown = errors.New("govim shutting down")
+)
+
+type errProto struct {
+	underlying error
+}
+
+type Transport interface {
+	Start() error
+	Close() error
+	Loaded() chan struct{}
+	Initialized() chan struct{}
+	IsShutdown() chan struct{}
+
+	Receive() (json.RawMessage, error)
+	Send(callback Callback, msgType string, params ...interface{}) error
+	SendJSON(p1, p2 interface{}, ps ...interface{})
+}
+
+type Callback struct{}
--- a/internal/transport/types.go
+++ b/internal/transport/types.go
@@ -20,8 +20,29 @@ type Transport interface {
 	Initialized() chan struct{}
 	IsShutdown() chan struct{}
 
-	Read() (int, json.RawMessage, error)
+	Read() (id int, messageType string, args []json.RawMessage, err error)
+	Send(callback Callback, callbackType string, params ...interface{}) error
 	SendJSON(p1, p2 interface{}, ps ...interface{})
 }
 
-type Callback struct{}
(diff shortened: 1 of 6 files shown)


internal/transport/types.go
```diff
@@ -0,0 +1,50 @@
...
+type errProto struct {
+	underlying error
+}
+
+type Transport interface {
+	Start() error
+	Close() error
+	Loaded() chan struct{}
+	Initialized() chan struct{}
+	IsShutdown() chan struct{}
```
milesbxf: These don't do anything at the moment - do we need them 🤔

internal/transport/types.go
```diff
@@ -0,0 +1,50 @@
...
+
+	Read() (id int, messageType string, args []json.RawMessage, err error)
+	Send(callback Callback, callbackType string, params ...interface{}) error
+	SendAndReceive(messageType string, args ...interface{}) (json.RawMessage, error)
+	SendAndReceiveAsync(messageType string, args ...interface{}) (ScheduledCallback, error)
+	SendJSON(p1, p2 interface{}, ps ...interface{})
```
milesbxf: Need to remove references to JSON here. Also think we should come up with a better abstraction than `SendAndReceive/SendAndReceiveAsync` - possibly Futures is a good concept to use for the latter (and make the former a helper function that's just sugar around it)

internal/transport/vim.go
```diff
@@ -0,0 +1,252 @@
...
+func NewVimTransport(in io.Reader, out io.Writer, log io.Writer, eventQueue *queue.Queue) Transport {
+	return &vimTransport{
+		in:  json.NewDecoder(in),
+		out: json.NewEncoder(out),
+		log: log,
+
+		callVimNextID: 1,
+		callbackResps: make(map[int]Callback),
+
+		eventQueue: eventQueue,
```
milesbxf: Replace this with a passed in callback that can enqueue things on the event queue, we really don't need the whole event queue

internal/transport/vim.go
```diff
@@ -0,0 +1,252 @@
...
+	return nil
+}
+
+func (v *vimTransport) run() error {
+	for {
+		select {
+		case <-v.tomb.Dying():
+			return tomb.ErrDying
+		}
+	}
```
milesbxf: At the moment this is purely just to keep the `tomb` alive, but I wonder if there's a better way of doing this. We might need this in future if we want to move more of the event loop into here

internal/transport/vim.go
```diff
@@ -0,0 +1,252 @@
...
+
+func (v *vimTransport) Initialized() chan struct{} {
+	return nil
+}
+
+func (v *vimTransport) IsShutdown() chan struct{} {
+	return nil
+}
+
+func (v *vimTransport) Read() (int, string, []json.RawMessage, error) {
```
milesbxf: Ideally we wouldn't need to return the Vim channel ID, and we would have marshalled `json.RawMessage` into a generic Go type

myitcv:
Apologies, this PR got automatically closed as a result of the changes for #910 - we should have changed the base of open PRs before deleting the `master` branch. Please can you change the base of this PR from `master` to `main`?

PR #346, opened 2019-07-07
cmd/govim: support batching of function calls to Vim

myitcv:


myitcv:
@milesbxf @leitzler - I'm minded to go with this approach to see how things work out, unless you have an immediate thoughts against it?

PR #366, opened 2019-07-10
[WIP] cmd/govim: use signs for diagnostics

### LINKED ISSUE #284: cmd/govim: add support for quickfix item signs
This is a proposal.

It would be nice if quickfix items could be annotated using vim's built in "sign".
There is a rough proof-of-concept implementation at https://github.com/leitzler/govim (master) that defines a new sign named "err" (ought to be something with "govim") using:
```
sign define err text=>> texthl=Error
```
It tells vim to place ">>" using the "Error" highlight when the sign named "err" is used.

And then each time the quickfix list is populated it removes all signs (regardless of buffer or file, can be improved) using:
```
sign unplaced *
```

.. to finally place a new sign next to each item:
```
sign place <Id> line=<lineno> name=err file=<filename>
```

Note that due to gopls dropping some diagnostics the sign placing won't be as consistent as it should.

leitzler:
Closes #284 

This PR will place one sign on each line where there is a quickfix entry.


myitcv:
FYI - https://github.com/vim/vim/commit/809ce4d317fe12db0b2c17f16b4f77200fb060c4 just landed.

leitzler:
Closed in favour of #382 that uses `sign_placelist()` and `sign_unplacelist()` to place/remove several signs at once.

PR #382, opened 2019-07-16
[WIP] cmd/govim: add signs to quickfix entries (#284)

### LINKED ISSUE #284: cmd/govim: add support for quickfix item signs

leitzler:
Closes #284 

This PR will place one sign on each line where there is a quickfix entry. It superseeds PR #366 since vim now supports placing and removing several signs in a single call and that works better than batching as far as I've seen.

### DIFF
--- a/cmd/govim/main.go
+++ b/cmd/govim/main.go
@@ -206,6 +206,11 @@ func (g *govimplugin) Init(gg govim.Govim, errCh chan error) error {
 	g.DefineFunction(string(config.FunctionSetUserBusy), []string{"isBusy"}, g.vimstate.setUserBusy)
 	g.DefineCommand(string(config.CommandReferences), g.vimstate.references)
 	g.DefineCommand(string(config.CommandRename), g.vimstate.rename, govim.NArgsZeroOrOne)
+	if g.placeSigns() {
+		if err := g.vimstate.signDefine(); err != nil {
+			return fmt.Errorf("failed to define signs: %v", err)
+		}
+	}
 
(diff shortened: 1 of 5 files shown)


leitzler:
(Requires vim 8.1.1682)

myitcv:
Will take a proper look later, but first, take a look at the following for ideas on how to conditionally do things based on version:

https://github.com/myitcv/govim/blob/0346d4df12cd35bfe9fca3923dc358cac113b824/cmd/govim/main.go#L344-L355

myitcv:
And you can do the same sort of constraints in testscript tests:

https://github.com/myitcv/govim/blob/0346d4df12cd35bfe9fca3923dc358cac113b824/cmd/govim/testdata/references.txt#L18-L19

leitzler:
See #390

PR #390, opened 2019-07-18
cmd/govim: add signs to quickfix entries (#284)

leitzler:
Moved branch due to CI issues, see #382

### DIFF
--- a/cmd/govim/signs.go
+++ b/cmd/govim/signs.go
@@ -31,6 +31,11 @@ func (v *vimstate) signDefine() error {
 	return nil
 }
 
+// getPlacedDict is the representation of arguments used in vim's sign_getplaced()
+type getPlacedDict struct {
+	Group string `json:"group"`
+}
+
 // bufferSigns represents a single element in the response from a sign_getplaced() call
 type bufferSigns struct {
 	Signs []struct {
@@ -41,19 +46,6 @@ type bufferSigns struct {
 		Group string `json:"group"`
 	} `json:"signs"`
 	BufNr int `json:"bufnr"`
-}
-
-// signGetPlaced returns all signs placed by govim in a specific buffer
-func (v *vimstate) signGetPlaced(buf int) (bufferSigns, error) {
-	argDict := struct {
-		Group string `json:"group"`
-	}{signGroup}
-
-	var out []bufferSigns
-	v.Parse(v.ChannelCall("sign_getplaced", buf, argDict), &out)
-	// out[0] should always exist and be the only element since sign_getplaced is
-	// called with a signle buffer as argument
-	return out[0], nil
 }
 
 // placeDict is the representation of arguments used in vim's sign_place() and sign_placelist()
@@ -83,9 +75,23 @@ func (v *vimstate) redefineSigns(fixes []quickfixEntry) error {
 	remove := make(map[bufLine]int) // Value is sign ID, used to unplace duplicates
 	place := make(map[bufLine]int)  // Value is insert order, used to avoid sorting
 
-	// Assume all existing signs should be removed, unless found in quickfix entry list
+	// One call per buffer is needed since sign_getplaced() doesn't support getting
+	// signs from all buffers within a specific sign group.
+	v.BatchStart()
 	for buf := range v.buffers {
-		placed, _ := v.signGetPlaced(buf)
+		v.BatchChannelCall("sign_getplaced", buf, getPlacedDict{signGroup})
+	}
+
+	bufs := make([]bufferSigns, 0)
+	for _, res := range v.BatchEnd() {
+		var tmp []bufferSigns
+		v.Parse(res, &tmp)
+		bufs = append(bufs, tmp...)
+	}
+
+	v.BatchStart()
+	// Assume all existing signs should be removed, unless found in quickfix entry list
+	for _, placed := range bufs {
 		for _, sign := range placed.Signs {
 			bl := bufLine{placed.BufNr, sign.Lnum}
 			if _, exist := remove[bl]; exist {
@@ -130,8 +136,7 @@ func (v *vimstate) redefineSigns(fixes []quickfixEntry) error {
 				Lnum:   bl.line,
 				Name:   errorSign}
 		}
-
-		v.ChannelCall("sign_placelist", placeList)
+		v.BatchChannelCall("sign_placelist", placeList)
 	}
 
 	// Remove signs on all lines that didn't have a corresponding quickfix entry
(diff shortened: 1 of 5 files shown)


varun06:
excited for this.

cmd/govim/signs.go
```diff
@@ -0,0 +1,148 @@
...
+	text          string // One or two chars shown in the gutter
+	textHighlight string // Highlight used
+}
+
+// Only one sign type at the moment, errors.
+var errorSign = signDef{"govimerr", ">>", "Error"}
+
+// signDefine defines the govim error sign and must be called once before placing any signs
+func (v *vimstate) signDefine() error {
+	argDict := struct {
```
myitcv: Why not use `signDef` instead of defining another type here?

cmd/govim/signs.go
```diff
@@ -0,0 +1,148 @@
...
+	} `json:"signs"`
+	BufNr int `json:"bufnr"`
+}
+
+// signGetPlaced returns all signs placed by govim in a specific buffer
+func (v *vimstate) signGetPlaced(buf int) (bufferSigns, error) {
+	argDict := struct {
+		Group string `json:"group"`
+	}{signGroup}
+	resp := v.ParseJSONArgSlice(v.ChannelCall("sign_getplaced", buf, argDict))
```
myitcv: Perhaps:

```
var out []bufferSigns
v.Parse(v.ChannelCall("sign_getplaced", buf, argDict), &out)
return out[0], nil
```

cmd/govim/signs.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// unplaceDict is the representation of arguments used in vim's sign_unplace() and sign_unplacelist()
+type unplaceDict struct {
+	Buffer int    `json:"buffer,omitempty"`
+	Group  string `json:"group,omitempty"` // sign_unplacelist() only
+	ID     int    `json:"id,omitempty"`
+}
+
+// redefineSigns ensures that there is only one govim sign per buffer and line
```
myitcv: > one govim sign per buffer line

Superfluous "and"?

testsetup/testsetup.go
```diff
@@ -24,6 +24,7 @@ const (
 	EnvLoadTestAPI               = "GOVIM_LOAD_TEST_API"
 	EnvDisableIncrementalSync    = "GOVIM_DISABLE_INCREMENTALSYNC"
 	EnvDisablePopupWindowBalloon = "GOVIM_DISABLE_POPUPWINDOWBALLOON"
+	EnvDisableSignPlace          = "GOVIM_DISABLE_SIGNPLACE"
```
myitcv: 👍

testsetup/testsetup.go
```diff
@@ -37,6 +38,10 @@ const (
 	// windows for balloons
 	MinPopupWindowBalloon = "v8.1.1657"
 
+	// MinSignPlace is the minimum version of Vim required to use
+	// sign placement (i.e. supports sign_placelist() / sign_unplacelist())
+	MinSignPlace = "v8.1.1682"
```
myitcv: 👍

cmd/govim/signs.go
```diff
@@ -0,0 +1,148 @@
...
+	for buf := range v.buffers {
+		placed, _ := v.signGetPlaced(buf)
+		for _, sign := range placed.Signs {
+			bl := bufLine{placed.BufNr, sign.Lnum}
+			if _, exist := remove[bl]; exist {
+				// As each sign isn't tracked individually, we might end up with several
+				// signs on the same line when, for example, a line is removed.
+				// By removing duplicates here we ensure that there is only one
+				// sign per line.
+				v.ChannelCall("sign_unplace", signGroup, unplaceDict{Buffer: bl.buf, ID: sign.ID})
```
myitcv: Any reason you're not using batching here and for the calls that follow?

myitcv:
Just reminding myself that once we land these changes, we should look to revisit https://github.com/myitcv/govim/issues/289. Setting a buffer variable with the error count will make it trivial to define, for example, a custom vim airline section/indicator to show if there are any errors in a buffer.

testsetup/testsetup.go
myitcv: Just to clarify, I'm giving this a 👍 because this is an immature new feature that we want to test over time with the setting of this env variable being a nuclear option in case people find issues.

cmd/govim/signs.go
leitzler: Leftovers from old implementation, fixed!

cmd/govim/signs.go
leitzler: Looks better, updated

cmd/govim/signs.go
leitzler: :+1:

cmd/govim/signs.go
leitzler: No, not really :). Added batching as well.

leitzler: Oh, turns out goimports can start a batch that isn't terminated before signs are redefined. When using batch to place signs `cmd/govim/testdata/goimports.txt` panics.

I reverted the change for now.

myitcv: This doesn't sound right. There shouldn't be two batches overlapping ever. 

Can you point me to a commit where this panics?

cmd/govim/signs.go
```diff
@@ -0,0 +1,153 @@
...
+	place := make(map[bufLine]int)  // Value is insert order, used to avoid sorting
+
+	// One call per buffer is needed since sign_getplaced() doesn't support getting
+	// signs from all buffers within a specific sign group.
+	v.BatchStart()
+	for buf := range v.buffers {
+		v.BatchChannelCall("sign_getplaced", buf, getPlacedDict{signGroup})
+	}
+
+	bufs := make([]bufferSigns, 0)
```
myitcv: Just a small retrospective point, I would always declare this as:

```go
var bufs []bufferSigns
```

cmd/govim/signs.go
```diff
@@ -0,0 +1,153 @@
...
+	// delete existing entries from the list of signs to removed
+	inx := 0
+	for _, f := range fixes {
+		bl := bufLine{f.Buf, f.Lnum}
+		if _, exist := remove[bl]; exist {
+			delete(remove, bl)
+			continue
+		}
+
+		if bl.buf == -1 {
```
myitcv: How do we end up with an unknown buffer here?

myitcv: Ah, ignore this question. It's totally clear how.

PR #398, opened 2019-07-22
cmd/govim: add support for motions

myitcv:


cmd/govim/motions.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, fmt.Errorf("got unknown target %q", target)
+	}
+
+	// Brute force this for now
+	var targetNode ast.Node
+	switch dir {
+	case "next":
+		for i := 0; i < len(b.AST.Decls); i++ {
+			d := b.AST.Decls[i]
+			v.Logf("Checking %v > %v\n", resolv(d), pos)
```
leitzler: Should this be here? If so, would it make sense to add it in "prev" as well for consistency?

cmd/govim/motions.go
```diff
@@ -0,0 +1,128 @@
...
+	default:
+		return nil, fmt.Errorf("got unknown direction %q", dir)
+	}
+	var resolv func(n ast.Node) token.Pos
+	switch target {
+	case "File.Decls.End()":
+		resolv = func(n ast.Node) token.Pos {
+			// The user sees themselves as being at the end when the cursor is
+			// before the closing brace, not after. Hence adjust backwards
+			offset := b.Fset.File(pos).Offset(n.End())
```
leitzler: This will panic if the last statement is broken, e.g.:
```go
package main

func main() {
```
Running `call GOVIMMotion("next", "File.Decls.End()")` in that file will return an end position outside the file.

cmd/govim/motions.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"golang.org/x/tools/go/ast/astutil"
+)
+
+func (v *vimstate) motion(args ...json.RawMessage) (interface{}, error) {
+	// GOVIMMotion has the signature:
+	//
+	//     func GOVIMMotion(direction, target string)
+	//
+	// direction is either "previous", "next", "inner" or "outer" (relative to
```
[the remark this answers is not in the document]
myitcv: Thanks, fixed. These comments will, I suspect, quickly become out of date as this concept of motions matures.

cmd/govim/motions.go
```diff
@@ -0,0 +1,128 @@
...
+	// the cursor position). target is based as closely as possible on the
+	// definitions in go/ast, the one current exception to this being Block whic
+	// is used to
```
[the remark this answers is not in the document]
myitcv: Thanks, removed. I didn't have anything particularly useful to say :)

cmd/govim/motions.go
myitcv: Removed, thanks

cmd/govim/motions.go
myitcv: Addressed, per Slack conversation.

PR #399, opened 2019-07-23
cmd/govim: add QuickfixSignsDisable config

### LINKED ISSUE #395: cmd/govim: add config option to disable placing of signs
We already have [`QuickfixAutoDiagnosticsDisable `](https://godoc.org/github.com/myitcv/govim/cmd/govim/config#Config); we should have an equivalent for people who want to see the quickfix being auto-populated but who do not want the signs automatically placed.

The default should remain that they are placed.

leitzler:
Closes #395

### DIFF
--- a/cmd/govim/config/config.go
+++ b/cmd/govim/config/config.go
@@ -23,6 +23,9 @@ type Config struct {
 
 	// QuickfixSignsDisable is a boolean (0 or 1 in VimScript) that controls
 	// whether quickfix entries should be annotated with signs in the gutter.
+	// Signs are placed when gopls diagnostics updates the quickfix list, either
+	// automatically when QuickfixAutoDiagnosticsDisable is false, or when the
+	// user run :GOVIMQuickfixDiagnostics.
 	// Default: false (0)
 	QuickfixSignsDisable bool
 
--- a/cmd/govim/config/config.go
+++ b/cmd/govim/config/config.go
@@ -21,6 +21,11 @@ type Config struct {
 	// population. Default: false (0)
 	QuickfixAutoDiagnosticsDisable bool
 
+	// QuickfixSignsDisable is a boolean (0 or 1 in VimScript) that controls
+	// whether quickfix entries should be annotated with signs in the gutter.
+	// Default: false (0)
+	QuickfixSignsDisable bool
+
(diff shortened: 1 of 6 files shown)


cmd/govim/config/config.go
```diff
@@ -21,6 +21,11 @@ type Config struct {
 	// population. Default: false (0)
 	QuickfixAutoDiagnosticsDisable bool
 
+	// QuickfixSignsDisable is a boolean (0 or 1 in VimScript) that controls
+	// whether quickfix entries should be annotated with signs in the gutter.
+	// Default: false (0)
```
myitcv: Only one more small comment: I think we need to mention that this setting only has impact if the auto diagnostics are not disabled.

cmd/govim/testdata/signs.txt
```diff
@@ -8,6 +8,9 @@
 [vim] [!vim:v8.1.1682] skip
 [gvim] [!gvim:v8.1.1682] skip
 
+# Ensure sign placment are enabled
+vim call 'govim#config#Set' '["QuickfixSignsDisable", 0]'
```
myitcv: Isn't the intention here to verify the default is disabled instead?

In which case, perhaps this line shifts to another testscript where we assert the value returned by `govim#config#Get()`? i.e. we create a testscript for asserting the out-of-the-box config for all settings? Thoughts?

cmd/govim/config/config.go
leitzler: I'll clarify the behaviour. It do actually have impact since it's possible to have signs enabled without the automatic diagnostics, i.e. placed when the user run `:GOVIMQuickfixDiagnostics`

cmd/govim/testdata/signs.txt
leitzler: Yeah, good point.
I thought about adding `govim#config#Get()` to `quickfix_config.txt`, but it isn't populated with default values if there are none (maybe it should?) so I just tested the default behaviour there.

I also had an idea of maybe splitting up the `quickfix_config.txt` into different versions since the version handling clutters the readability. Then it would be easy to drop, for example, `quickfix_config_pre_v8.1.1682` if that version isn't supported anymore. But I decided to leave it as is for now.

I'll remove the call from `signs.txt` as solution for this .

PR #400, opened 2019-07-24
scripts: fix GOPATH in runDockerRun.sh

leitzler:
Go doesn't require the env var GOPATH to be set since 1.8, `go env GOPATH` will however report the correct one even if it isn't set.

### DIFF
--- a/_scripts/runDockerRun.sh
+++ b/_scripts/runDockerRun.sh
@@ -10,7 +10,7 @@ proxy=""
 
 if [ "${CI:-}" != "true" ]
 then
-	proxy="-v $GOPATH/pkg/mod/cache/download:/cache -e GOPROXY=file:///cache"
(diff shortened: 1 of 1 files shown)


_scripts/runDockerRun.sh
```diff
@@ -10,7 +10,7 @@ proxy=""
 
 if [ "${CI:-}" != "true" ]
 then
-	proxy="-v $GOPATH/pkg/mod/cache/download:/cache -e GOPROXY=file:///cache"
+	proxy="-v $(go env GOPATH)/pkg/mod/cache/download:/cache -e GOPROXY=file:///cache"
```
myitcv: Thanks - good catch. 

Although to be 100% correct I think you need just the first element:

```
go env GOPATH | sed -e 's/:/\n/' | head -n 1
```

leitzler: Ah, good point!

PR #413, opened 2019-07-29
cmd/govim: add signcolumn=yes to minimal vimrc

myitcv:
The out-of-the-box experience should be that the sign column is visible,
so as to prevent the dance of it (the sign column)
appearing/disappearing as errors come and go.


myitcv:
Known test flakes. Merging.

Issue #433, opened 2019-07-31
testdriver: keep post-match content after errlogmatch


leitzler:
When `errlogmatch` is used in the testscripts twice the second match will only be able to find things that has been added since the last call.

It would be a better default to keep everything after the last match instead, otherwise we might end up throwing away content that we actually want to match that happened to be written before the first call.

Consider the following case with a pseudo log like this:
```
Entry1
Entry2
Entry3
// call errlogmatch Entry1 here
Entry4
Entry5
// call errlogmatch Entry3 here
```

The second call won't match since the first "discards" everything before the call.
A first quick hack shows that changing this behavior won't break any existing tests.

leitzler:
One thing to decide before implementing this, how should the "search from start" flag be handled?

Example:
```
Entry1
Entry2
Entry3
// call errlogmatch Entry2 here, next call will be able to find anything AFTER Entry2
Entry4
Entry5
// call errlogmatch Entry1 here, with "search from start" enabled
Entry6
```

Should the next call start searching from `Entry2` (I.e. move start position backwards), `Entry6` (move start pos forward) or even `Entry3` (not update start pos)?

myitcv:
"Search from start" is exactly that, no? From the very start of the log.

leitzler:
Sorry for being unclear, I meant the next call to `errlogmatch` after a "search from start" call.
As per Slack conversation - a "search from start" call should update and potentially move the start position for next match backwards.

PR #435, opened 2019-07-31
testdriver: update errlogmatch logic

leitzler:
This PR will update `errlogmatch` so that the search start offset for the next call is based on the last match instead of how the errlog looked at the last call.

It will also rename the flag `leavelast` to `peek` to reflect the new behavior.


leitzler:
Fixes #433

PR #477, opened 2019-09-02
cmd/govim: add separate watcher for darwin

### LINKED ISSUE #464: cmd/govim: use of fsnotify holds too many open files on macOS
### What version of Vim/Gvim are you using?

<details><summary>Vim/Gvim version</summary><br><pre>
$ vim --version
VIM - Vi IMproved 8.1 (2018 May 18, compiled Jul  5 2019 16:04:22)
macOS version
Included patches: 1-1518
Compiled by pauljolly@Pauls-MacBook-Pro.local
Huge version without GUI.  Features included (+) or not (-):
+acl               -farsi             -mouse_sysmouse    -tag_any_white
+arabic            +file_in_path      +mouse_urxvt       -tcl
+autocmd           +find_in_path      +mouse_xterm       +termguicolors
+autochdir         +float             +multi_byte        +terminal
-autoservername    +folding           +multi_lang        +terminfo
-balloon_eval      -footer            -mzscheme          +termresponse
+balloon_eval_term +fork()            +netbeans_intg     +textobjects
-browse            -gettext           +num64             +textprop
++builtin_terms    -hangul_input      +packages          +timers
+byte_offset       +iconv             +path_extra        +title
+channel           +insert_expand     -perl              -toolbar
+cindent           +job               +persistent_undo   +user_commands
-clientserver      +jumplist          +postscript        +vartabs
+clipboard         +keymap            +printer           +vertsplit
+cmdline_compl     +lambda            +profile           +virtualedit
+cmdline_hist      +langmap           -python            +visual
+cmdline_info      +libcall           -python3           +visualextra
+comments          +linebreak         +quickfix          +viminfo
+conceal           +lispindent        +reltime           +vreplace
+cryptv            +listcmds          +rightleft         +wildignore
-cscope            +localmap          -ruby              +wildmenu
+cursorbind        -lua               +scrollbind        +windows
+cursorshape       +menu              +signs             +writebackup
+dialog_con        +mksession         +smartindent       -X11
+diff              +modify_fname      -sound             -xfontset
+digraphs          +mouse             +spell             -xim
-dnd               -mouseshape        +startuptime       -xpm
-ebcdic            +mouse_dec         +statusline        -xsmp
+emacs_tags        -mouse_gpm         -sun_workshop      -xterm_clipboard
+eval              -mouse_jsbterm     +syntax            -xterm_save
+ex_extra          +mouse_netterm     +tag_binary        
+extra_search      +mouse_sgr         -tag_old_static    
   system vimrc file: "$VIM/vimrc"
     user vimrc file: "$HOME/.vimrc"
 2nd user vimrc file: "~/.vim/vimrc"
      user exrc file: "$HOME/.exrc"
       defaults file: "$VIMRUNTIME/defaults.vim"
  fall-back for $VIM: "/Users/pauljolly/usr/vim/share/vim"
Compilation: gcc -c -I. -Iproto -DHAVE_CONFIG_H   -DMACOS_X -DMACOS_X_DARWIN  -g -O2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1       
Linking: gcc   -L/usr/local/lib -o vim        -lm -lncurses  -liconv -framework AppKit           

</pre></details>

<pre>
$ go version
go version go1.12.6 darwin/amd64
</pre>

$ go env
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/pauljolly/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/pauljolly/gostuff"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/Cellar/go/1.12.6/libexec"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.12.6/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/dr/dr09wv9n7l9_j_z635q0p7rr0000gp/T/go-build502338012=/tmp/go-build -gno-record-gcc-switches -fno-common"
</pre></details>

b6bbbaec1c8457db55c5f7e26d3030fe5c118d13

### What is the output of `:messages` in Vim?

n/a

### Do the Vim channel, `govim` or `gopls` log files show anything interesting?

n/a

### How did you install `govim`?

n/a

### What did you do?

Open `govim.go` in a clone of this repo

A small number of open files reported by `lsof` for the `govim` process

### What did you see instead?

~255 files open, i.e. all of the files in the govim repo. These files are open because `fsnotify` uses a slightly old mechanism for watching files on macOS: https://github.com/fsnotify/fsnotify/issues/11

leitzler:
As the current filesystem watcher is using `fsnotify/fsnotify` and opens up loads of files this PR will change the macOS implementation to use `fsnotify/fsevents` instead.

Pro's about this is that `fsevents` watches recursively so the implementation is fairly simple.

Con's about this is that it doesn't support module directories that spawns over several devices as the device must be specified when the watcher is created. There is an open issue, https://github.com/fsnotify/fsevents/issues/36, that prevents adding multiple paths.

I guess we could add several watchers to handle that case, but I'm not sure it's worth it as the watchers should go away as soon as they are added to gopls anyway.

Another option could be to allow macOS users to use the "legacy fsnotifiy" with an env var, in case they run into problems with fsevents.

Closes #464

### DIFF
--- a/cmd/govim/watcher.go
+++ b/cmd/govim/watcher.go
@@ -8,8 +8,8 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/fsnotify/fsnotify"
 	"github.com/myitcv/govim"
+	"github.com/myitcv/govim/cmd/govim/internal/fswatcher"
 	"github.com/myitcv/govim/cmd/govim/internal/golang_org_x_tools/lsp/protocol"
 	"github.com/myitcv/govim/cmd/govim/internal/golang_org_x_tools/span"
 	"github.com/myitcv/govim/cmd/govim/internal/types"
@@ -22,7 +22,7 @@ type modWatcher struct {
 	// disappear
 	*govimplugin
 
-	watcher *fsnotify.Watcher
+	watcher *fswatcher.FSWatcher
 
 	// root is the directory root of the watch
 	root string
@@ -35,12 +35,14 @@ type modWatcher struct {
 	files map[string]int
 }
 
+func (mw *modWatcher) close() error { return mw.watcher.Close() }
+
 // newWatcher returns a new watcher that will "watch" on the Go files in the
 // module identified by gomodpath
 func newModWatcher(plug *govimplugin, gomodpath string) (*modWatcher, error) {
-	mw, err := fsnotify.NewWatcher()
+	w, err := fswatcher.New(gomodpath, &plug.tomb)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create new watcher: %v", err)
+		return nil, err
 	}
 
 	dirpath := filepath.Dir(gomodpath)
@@ -51,7 +53,7 @@ func newModWatcher(plug *govimplugin, gomodpath string) (*modWatcher, error) {
 
 	res := &modWatcher{
 		govimplugin: plug,
-		watcher:     mw,
+		watcher:     w,
 		root:        dirpath,
 		watches:     make(map[string]bool),
 		files:       make(map[string]int),
@@ -59,16 +61,12 @@ func newModWatcher(plug *govimplugin, gomodpath string) (*modWatcher, error) {
 
 	go res.watch()
 	// fake event to kick start the watching
-	mw.Events <- fsnotify.Event{
-		Name: dirpath,
-		Op:   fsnotify.Create,
+	res.watcher.Events() <- fswatcher.Event{
+		Path: dirpath,
+		Op:   fswatcher.OpChanged,
 	}
 
 	return res, nil
-}
-
-func (m *modWatcher) close() error {
-	return m.watcher.Close()
 }
 
 func (m *modWatcher) watch() {
@@ -78,19 +76,22 @@ func (m *modWatcher) watch() {
 	infof := func(format string, args ...interface{}) {
 		m.Logf("file watcher event: "+format, args...)
 	}
+	eventCh := m.watcher.Events()
+	errCh := m.watcher.Errors()
+
 	for {
 		select {
-		case event, ok := <-m.watcher.Events:
+		case event, ok := <-eventCh:
 			if !ok {
 				// watcher has been stopped?
 				return
 			}
 			switch event.Op {
-			case fsnotify.Remove, fsnotify.Rename:
-				path := event.Name
+			case fswatcher.OpRemoved:
+				path := event.Path
 				var didFind bool
 				for ew := range m.watches {
-					if event.Name == ew || strings.HasPrefix(ew, event.Name+string(os.PathSeparator)) {
+					if event.Path == ew || strings.HasPrefix(ew, event.Path+string(os.PathSeparator)) {
 						didFind = true
 						if err := m.watcher.Remove(ew); err != nil {
 							errf("failed to remove watch on %v: %v", ew, err)
@@ -108,8 +109,8 @@ func (m *modWatcher) watch() {
 				m.Schedule(func(govim.Govim) error {
 					return m.vimstate.handleEvent(event)
 				})
-			case fsnotify.Create, fsnotify.Write, fsnotify.Chmod:
-				path := event.Name
+			case fswatcher.OpChanged:
+				path := event.Path
 				dirInfo, err := os.Stat(path)
 				if err != nil {
 					errf("failed to stat %v: %v", path, err)
@@ -128,7 +129,7 @@ func (m *modWatcher) watch() {
 				}
 
 				// Walk the dir that is event.Name
-				err = filepath.Walk(event.Name, func(path string, info os.FileInfo, err error) error {
+				err = filepath.Walk(event.Path, func(path string, info os.FileInfo, err error) error {
 					if err != nil {
 						return err
 					}
@@ -154,10 +155,10 @@ func (m *modWatcher) watch() {
 					return err
 				})
 				if err != nil {
-					errf("failed to walk %v: %v", event.Name, err)
+					errf("failed to walk %v: %v", event.Path, err)
 				}
 			}
-		case err, ok := <-m.watcher.Errors:
+		case err, ok := <-errCh:
 			if !ok {
 				// watcher has been stopped?
 				return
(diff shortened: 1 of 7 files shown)


cmd/govim/oswatcher.go
```diff
@@ -0,0 +1,47 @@
+// +build !darwin
+
+package main
```
myitcv: I'd be half tempted to move this to an `internal` package now that things are growing a bit. 

That will also help to isolate what we're at some point going to throw away, too, keeping it separate from the code we will keep

cmd/govim/oswatcher.go
```diff
@@ -0,0 +1,47 @@
...
+			case fsnotify.Rename, fsnotify.Remove:
+				eventCh <- fsEvent{e.Name, fsOpRemoved}
+			case fsnotify.Create, fsnotify.Chmod, fsnotify.Write:
+				eventCh <- fsEvent{e.Name, fsOpChanged}
+			}
+		}
+		close(eventCh)
+	}()
+
+	return &osWatcher{
```
myitcv: This is all nicely readable and clean, but I think I'd prefer to define an interface and return a value of that type from `NewOSWatcher` (assuming we put this in an internal package)

cmd/govim/oswatcher.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func newOSWatcher(gomodpath string) (*osWatcher, error) {
+	mw, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new watcher: %v", err)
+	}
+
+	eventCh := make(chan fsEvent)
+	go func() {
```
myitcv: We should pass a `*gopkg.in/tomb.v2.tomb.Tomb` to `NewOSWatcher` (assuming we put this in an internal package) and then spawn this go routine using `*gopkg.in/tomb.v2.tomb.Tomb.Go()`. 

The approach is not perfect, but it's helped track down errors in the past

cmd/govim/oswatcher_darwin.go
```diff
@@ -0,0 +1,63 @@
...
+	eventCh := make(chan fsEvent)
+	go func() {
+		for {
+			events, ok := <-es.Events
+			if !ok {
+				break
+			}
+			for i := range events {
+				event := events[i]
+				path := filepath.Join("/", event.Path)
```
myitcv: What's going on here?

cmd/govim/oswatcher_darwin.go
```diff
@@ -0,0 +1,63 @@
...
+func newOSWatcher(gomodpath string) (*osWatcher, error) {
+	dirpath := filepath.Dir(gomodpath)
+	dev, err := fsevents.DeviceForPath(dirpath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve device for path %v: %v", dirpath, err)
+	}
+
+	es := &fsevents.EventStream{
+		Paths:   []string{dirpath},
+		Latency: 200 * time.Millisecond,
```
myitcv: This is, I'm guessing a throttle on the events? Why set any value?

cmd/govim/oswatcher_darwin.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/fsnotify/fsevents"
+)
+
+const fRemoved = fsevents.ItemRemoved | fsevents.ItemRenamed
+const fChanged = fsevents.ItemCreated | fsevents.ItemModified | fsevents.ItemChangeOwner
+
+func newOSWatcher(gomodpath string) (*osWatcher, error) {
+	dirpath := filepath.Dir(gomodpath)
+	dev, err := fsevents.DeviceForPath(dirpath)
```
myitcv: I'm totally fine with the restriction you describe in the PR. If people run into problems, let's fix them then

cmd/govim/watcher.go
```diff
@@ -35,12 +34,35 @@ type modWatcher struct {
 	files map[string]int
 }
 
+func (mw *modWatcher) close() error { return mw.watcher.close() }
+
+// osWatcher is a platform specific file watcher
+type osWatcher struct {
```
myitcv: Yeh, moving this stuff to an internal package feels like the right thing to do.

cmd/govim/watcher.go
```diff
@@ -80,19 +98,19 @@ func (m *modWatcher) watch() {
...
+		case event, ok := <-m.watcher.events:
 			if !ok {
 				// watcher has been stopped?
 				return
 			}
-			switch event.Op {
-			case fsnotify.Remove, fsnotify.Rename:
-				path := event.Name
+			switch event.op {
+			case fsOpRemoved:
```
myitcv: Just so you're aware: https://github.com/myitcv/govim/issues/465

cmd/govim/oswatcher.go
leitzler: The entire thing? I.e. both `watcher.go` and `oswatcher*.go` (terrible name, feel free to suggest a better if you come up with one).

cmd/govim/oswatcher_darwin.go
leitzler: From the [docs](https://developer.apple.com/documentation/coreservices/1447341-fseventstreamcreaterelativetodev):
> **latency**
The number of seconds the service should wait after hearing about an event from the kernel before passing it along to the client via its callback. Specifying a larger value may result in more effective temporal coalescing, resulting in fewer callbacks.

So I went for an arbitrary value. The example in their repo uses 500ms. Setting it (close) to zero could generate lots of callbacks and unnecessary overhead. But we might go for a lower value and increase it if anyone experience issues?

cmd/govim/watcher.go
leitzler: Ah! I could add a few tests as well. Both for deleting and tests that proves that adding/removing sub-directories with files (since the different implementations handle sub dirs differently).

cmd/govim/oswatcher.go
leitzler: Ok! - I'll add tomb

cmd/govim/oswatcher.go
leitzler: I did consider it but went for this because it went for this instead since it is easier to trace back (especially since the two different versions are separated by build tag, so there is essentially only one). And the whole "accept interfaces & return concrete types".

That's at least the reasoning behind it, I'll try it out using an interface instead and see how that ends up!

cmd/govim/oswatcher_darwin.go
leitzler: Oh, sloppy leftover. It should be the actual mount point, because the event paths are relative to the mount point. It uses https://developer.apple.com/documentation/coreservices/1447341-fseventstreamcreaterelativetodev underneath.

I'll add something that resolves the mount root and use that instead.

cmd/govim/oswatcher_darwin.go
myitcv: Sounds good.

cmd/govim/watcher.go
myitcv: Deleting files currently has no effect in the version of `gopls` we're using, so let's leave those until later.

PR #487, opened 2019-09-04
deps: x/tools and gopls adb45749

myitcv:


cmd/govim/gopls_client.go
```diff
@@ -18,8 +18,9 @@ const (
 
 var _ protocol.Client = (*govimplugin)(nil)
 
-func (g *govimplugin) ShowMessage(context.Context, *protocol.ShowMessageParams) error {
-	panic("not implemented yet")
+func (g *govimplugin) ShowMessage(ctxt context.Context, params *protocol.ShowMessageParams) error {
+	g.logGoplsClientf("Asked to show a message: %v", params.Message)
```
myitcv: See https://github.com/myitcv/govim/issues/445

cmd/govim/gopls_client.go
```diff
@@ -46,19 +47,24 @@ func (g *govimplugin) WorkspaceFolders(context.Context) ([]protocol.WorkspaceFol
 func (g *govimplugin) Configuration(ctxt context.Context, params *protocol.ConfigurationParams) ([]interface{}, error) {
 	defer absorbShutdownErr()
 	g.logGoplsClientf("Configuration: %v", pretty.Sprint(params))
-	// Assert based on the current behaviour of gopls
-	want := 1
-	if got := len(params.Items); want != got {
-		return nil, fmt.Errorf("govim gopls client: expected %v item(s) in params; got %v", want, got)
+
+	// gopls now sends params.Items for each of the configured
+	// workspaces. For now, we assume that the first item will be
```
leitzler: Is this a safe assumption? We could iterate over them and respond with empty configs for all except `gopls` (until we know how multiple workspaces should work)

myitcv: Per our Slack chat, yes, this is a valid assumption for now, per @ianthehat. If the assumption is broken at any point, the code below will "trip".

PR #497, opened 2019-09-05
deps: upgrade to golpls and x/tools 59228eac

myitcv:
* internal/lsp: avoid invalid state due to context cancelation 59228eac


myitcv:
Not really a contentious change here so merging.

Issue #496, opened 2019-09-05
cmd/govim: investigate watcher test flake


myitcv:
Observed in https://travis-ci.org/myitcv/govim/jobs/581425669

```
--- FAIL: TestScripts (159.98s)
    --- FAIL: TestScripts/scripts (0.01s)
        --- FAIL: TestScripts/scripts/watcher (38.89s)
            testscript.go:308: 
                # Test that the file watcher picks up file changes that occurs outside the editor
                # New file in the same package (37.769s)
                > vim ex 'e main.go'
                > errlogmatch -wait 30s 'PublishDiagnostics callback: &protocol.PublishDiagnosticsParams{\n\S+:\s+URI:\s+"file://'$WORK/main.go
                > cp const.go.orig const.go
                > errlogmatch -wait 30s '&protocol.DidOpenTextDocumentParams{\n\S+:\s+TextDocument: protocol.TextDocumentItem{URI:"file://'$WORK/const.go
                > errlogmatch -wait 30s 'PublishDiagnostics callback: &protocol.PublishDiagnosticsParams{\n\S+:\s+URI:\s+"file://'$WORK/const.go
                > vim ex 'call cursor(6,16)'
                > vim ex 'call feedkeys(\"iConst2\\<ESC>\", \"x\")'
                > vim ex 'w'
                > errlogmatch -wait 30s 'PublishDiagnostics callback: &protocol.PublishDiagnosticsParams{\n\S+:\s+URI:\s+"file://'$WORK/main.go
                FAIL: testdata/watcher.txt:12: errlogmatch failed to find match
```

Full build log:

https://gist.github.com/myitcv/7956c76aec8d0820d1dc241e3b1be206

myitcv:
Seeing this quite a lot...

myitcv:
e.g. two fails in https://travis-ci.org/myitcv/govim/builds/581439455

leitzler:
This one could be related to an old Vim version. After bumping the minimum required Vim version I've run the entire suite 6 times (*8 jobs) without seeing a single watcher failure.

Let's monitor for a little while and then close this if it doesn't happen soon again.

leitzler:
Haven't seen since and happened quite often before. Closing it for now.

PR #499, opened 2019-09-06
cmd/govim: fix incorrect motion mapping

myitcv:
```
Currently we have ]] mapped to:

:call GOVIMMotion("next", "File.Decls.End()")

and ][ mapped to:

:call GOVIMMotion("next", "File.Decls.Pos()")

But this is the inverse of the Vim-default mappings we are looking to
"emulate":

]]  [count] sections forward or to the next '{' in the
    first column.  When used after an operator, then also
    stops below a '}' in the first column.  exclusive
    Note that exclusive-linewise often applies.

][  [count] sections forward or to the next '}' in the
    first column.  |exclusive|
    Note that |exclusive-linewise| often applies.

Fix this by invertig our mapping.
```

### DIFF
--- a/cmd/govim/testdata/motion.txt
+++ b/cmd/govim/testdata/motion.txt
@@ -3,12 +3,12 @@
 vim ex 'e main.go'
 
 # Next start of File.Decl
-vim ex 'normal ]['
+vim ex 'normal ]]'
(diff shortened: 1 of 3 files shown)


myitcv:
Ignorning test flake (which is covered by #496

Issue #424, opened 2019-07-29
cmd/govim: investigate gofmt test flake


myitcv:
Found whilst running `master` build  227ae10 

```
        --- FAIL: TestScripts/scripts/gofmt (30.76s)
            testscript.go:308: 
                # Test that govim#config#Set("FormatOnSave", "gofmt") and GOVIMGoFmt work
                # Notice below: file.go and file.go.orig are identical. This is so that we
                # don't need to copy file.go.orig to file.go before opening file.go Because
                # doing so triggers govim to consider the file change from a file watcher
                # perspective which throws the observed diagnostics etc.
                # :GOVIMGoFmt whole file (0.303s)
                # Format on save (0.069s)
                # Format on save (bad syntax) (29.597s)
                > cp file.go.bad file.go
                > vim ex 'e! file.go'
                > errlogmatch -wait 30s 'sendJSONMsg: \[0,\[\d+,"call","setqflist".*'
                FAIL: testdata/gofmt.txt:30: errlogmatch failed to find match
```

Full build log

https://gist.github.com/myitcv/61da358a04d99a949a3a58e8b956003e

myitcv:
We haven't seen this in a while so closing.

PR #500, opened 2019-09-06
all: update remaining references from myitcv/govim to govim/govim

myitcv:


myitcv:
Merging, because the failure is a known flake #424

Issue #466, opened 2019-08-24
cmd/govim: goimports fails in github.com/eliasnaur/gophercon-uk-2019-talk


eliasnaur:
### What version of Vim/Gvim are you using?

<details><summary>Vim/Gvim version</summary><br><pre>
$ vim --version
VIM - Vi IMproved 8.1 (2018 May 18, compiled Aug 22 2019 04:01:19)
macOS version
Included patches: 1-1900
Compiled by Homebrew
Huge version without GUI.  Features included (+) or not (-):
+acl               -farsi             -mouse_sysmouse    -tag_any_white
+arabic            +file_in_path      +mouse_urxvt       -tcl
+autocmd           +find_in_path      +mouse_xterm       +termguicolors
+autochdir         +float             +multi_byte        +terminal
-autoservername    +folding           +multi_lang        +terminfo
-balloon_eval      -footer            -mzscheme          +termresponse
+balloon_eval_term +fork()            +netbeans_intg     +textobjects
-browse            +gettext           +num64             +textprop
++builtin_terms    -hangul_input      +packages          +timers
+byte_offset       +iconv             +path_extra        +title
+channel           +insert_expand     +perl              -toolbar
+cindent           +job               +persistent_undo   +user_commands
-clientserver      +jumplist          +postscript        +vartabs
+clipboard         +keymap            +printer           +vertsplit
+cmdline_compl     +lambda            +profile           +virtualedit
+cmdline_hist      +langmap           -python            +visual
+cmdline_info      +libcall           +python3           +visualextra
+comments          +linebreak         +quickfix          +viminfo
+conceal           +lispindent        +reltime           +vreplace
+cryptv            +listcmds          +rightleft         +wildignore
+cscope            +localmap          +ruby              +wildmenu
+cursorbind        +lua               +scrollbind        +windows
+cursorshape       +menu              +signs             +writebackup
+dialog_con        +mksession         +smartindent       -X11
+diff              +modify_fname      -sound             -xfontset
+digraphs          +mouse             +spell             -xim
-dnd               -mouseshape        +startuptime       -xpm
-ebcdic            +mouse_dec         +statusline        -xsmp
+emacs_tags        -mouse_gpm         -sun_workshop      -xterm_clipboard
+eval              -mouse_jsbterm     +syntax            -xterm_save
+ex_extra          +mouse_netterm     +tag_binary
+extra_search      +mouse_sgr         -tag_old_static
   system vimrc file: "$VIM/vimrc"
[...]

PR #506, opened 2019-09-07
test: unset GOPATH before test

### LINKED ISSUE #472: cmd/govim: changes tests to run with GOPATH unset
On the back of @leitzler's excellent investigation into #466, it appears the bug being fixed in https://go-review.googlesource.com/c/tools/+/191600/ is uncovered when GOPATH is _not_ set. Given this is, today, the most likely mode in which people will run `gopls`, we should flip to this mode for all tests. If there are any edge cases we then particularly want to test with `GOPATH` set we can create tests as required.

leitzler:
To be able to detect issues with default GOPATH (like in https://github.com/govim/govim/issues/466), the tests should run w/o GOPATH set. This change un-exports it before running the tests.

Closes #472

### DIFF
--- a/_scripts/dockerRun.sh
+++ b/_scripts/dockerRun.sh
@@ -23,6 +23,9 @@ fi
 go version
 vim --version
 
+# ensure GOPATH isn't set, to detect issues like https://golang.org/issues/33918. Also see govim/#472.
+export -n GOPATH
+
 ./_scripts/revendorToolsInternal.sh
 
 go install golang.org/x/tools/gopls


myitcv:
I just did a quick test here of the following script in `cmd/govim/testdata`:

```
env GOPATH
exec go env GOPATH
```

and got:

```
                > env GOPATH
                GOPATH=
                > exec go env GOPATH
                [stdout]
                $WORK/home/go
```

So I think we're actually fine here. It's just that our tests were, for some reason, not simulating the conditions/packages/whatever that Elias was seeing. 

I think we can therefore close this and #472, particularly because the upgrade of `gopls` and `x/tools` has helped fix the original issue.

leitzler:
Yes, looks like his case isn't covered with current tests.
Agree that this should be closed, along with #472.

Issue #445, opened 2019-08-12
cmd/govim: properly handle LSP client ShowMessage


myitcv:
In https://github.com/myitcv/govim/pull/467 we simply log the fact that the server asked us (the client) to show a message. If the message is important we should show it, preferably as a popup.

dragfire:
Hi @myitcv, I would like to work on this. It would be very helpful if you could push me in the right direction. How do you decide if a message is important or not?

myitcv:
Thanks for the interest @dragfire 

> It would be very helpful if you could push me in the right direction

Very happy to. First thing to say: I think it's probably worth waiting for https://github.com/myitcv/govim/pull/467 to be merged before tackling this. Because it's the upgrade of `gopls` that causes these calls to be made.

> How do you decide if a message is important or not?

But you can, by all means, have a poke around. Taking a look at the client method:

https://github.com/myitcv/govim/blob/956c1ebcc75bed45653d44f26109d9bca1a772ad/cmd/govim/gopls_client.go#L21-L23

the second parameter has a type defined as follows:

https://github.com/golang/tools/blob/master/internal/lsp/protocol/tsprotocol.go#L1731-L1742

The `Type` field is in effect an enumeration:

https://github.com/golang/tools/blob/master/internal/lsp/protocol/tsprotocol.go#L3579-L3597

So in the first instance I think we want to show a notification popup when we receive a message of `Error` type.

What do I mean by a notification popup? I've proposed a potential approach in https://github.com/myitcv/govim/issues/318#issuecomment-513582991. That results in something like the following:

![Screen Shot 2019-09-04 at 16 06 37](https://user-images.githubusercontent.com/3374574/64267310-163c3300-cf2e-11e9-8da0-7b79c7978986.png)

See `:help popup_create-arguments` for more popup options.

Please feel free to ask any questions you like. Having new contributors helps to ensure that our docs (or lack thereof right now) for (new) contributors are up-to-date, sufficient etc.

dragfire:
@myitcv Thank you so much. I will play around with this a bit. Looking forward to contributing to this repo.

myitcv:
We just merged https://github.com/myitcv/govim/pull/487, so please feel free to have a play around against 6961210ff30a5862594ff04dde9f3a5f62120ae4

dragfire:
[...]

PR #508, opened 2019-09-07
cmd/govim: show vim popup when MessageType is Error in ShowMessage 

dragfire:
PR for #445

### DIFF
--- a/cmd/govim/gopls_client.go
+++ b/cmd/govim/gopls_client.go
@@ -19,28 +19,52 @@ const (
 var _ protocol.Client = (*govimplugin)(nil)
 
 func (g *govimplugin) ShowMessage(ctxt context.Context, params *protocol.ShowMessageParams) error {
-	g.logGoplsClientf("Asked to show a message: %v", params.Message)
+	g.logGoplsClientf("ShowMessage callback: %v", params.Message)
+	if params.Type != protocol.Error {
+		return nil
+	}
+
+	g.Schedule(func(govim.Govim) error {
+		opts := make(map[string]interface{})
+		opts["mousemoved"] = "any"
+		opts["moved"] = "any"
+		opts["padding"] = []int{0, 1, 0, 1}
+		opts["wrap"] = false
+		opts["border"] = []int{}
+		opts["highlight"] = "ErrorMsg"
+		opts["line"] = 1
+		opts["close"] = "click"
+
+		g.ChannelCall("popup_create", params.Message, opts)
+		return nil
+	})
 	return nil
 }
+
 func (g *govimplugin) ShowMessageRequest(context.Context, *protocol.ShowMessageRequestParams) (*protocol.MessageActionItem, error) {
 	panic("not implemented yet")
 }
+
 func (g *govimplugin) LogMessage(ctxt context.Context, params *protocol.LogMessageParams) error {
 	defer absorbShutdownErr()
 	g.logGoplsClientf("LogMessage callback: %v", pretty.Sprint(params))
 	return nil
 }
+
 func (g *govimplugin) Telemetry(context.Context, interface{}) error {
 	panic("not implemented yet")
 }
+
 func (g *govimplugin) RegisterCapability(ctxt context.Context, params *protocol.RegistrationParams) error {
 	defer absorbShutdownErr()
 	g.logGoplsClientf("RegisterCapability: %v", pretty.Sprint(params))
 	return nil
 }
+
 func (g *govimplugin) UnregisterCapability(context.Context, *protocol.UnregistrationParams) error {
 	panic("not implemented yet")
 }
+
 func (g *govimplugin) WorkspaceFolders(context.Context) ([]protocol.WorkspaceFolder, error) {
 	panic("not implemented yet")
 }
@@ -61,9 +85,11 @@ func (g *govimplugin) Configuration(ctxt context.Context, params *protocol.Param
 	res[0] = conf
 	return res, nil
 }
+
 func (g *govimplugin) ApplyEdit(context.Context, *protocol.ApplyWorkspaceEditParams) (*protocol.ApplyWorkspaceEditResponse, error) {
 	panic("not implemented yet")
 }
+
 func (g *govimplugin) Event(context.Context, *interface{}) error {
 	panic("not implemented yet")
 }
--- a/cmd/govim/test_config.go
+++ b/cmd/govim/test_config.go
@@ -1,12 +1,14 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"sort"
 	"strings"
 
 	"github.com/govim/govim"
 	"github.com/govim/govim/cmd/govim/config"
+	"github.com/govim/govim/cmd/govim/internal/golang_org_x_tools/lsp/protocol"
 )
 
 // This file contains config that would otherwise be in the
@@ -25,6 +27,7 @@ const (
 	FunctionBadBatch            config.Function = "BadBatch"
 	FunctionAssertFailedBatch   config.Function = "AssertFailedBatch"
 	FunctionNonBatchCallInBatch config.Function = "NonBatchCallInBatch"
+	FunctionShowMessagePopup    config.Function = config.InternalFunctionPrefix + "ShowMessagePopup"
 )
 
 func (g *govimplugin) InitTestAPI() {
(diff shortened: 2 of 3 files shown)


dragfire:
@myitcv please take a look

cmd/govim/gopls_client.go
```diff
@@ -20,30 +20,47 @@ var _ protocol.Client = (*govimplugin)(nil)
...
 	g.logGoplsClientf("Asked to show a message: %v", params.Message)
+	if params.Type == protocol.Error {
+		opts := make(map[string]interface{})
+		opts["mousemoved"] = "any"
+		opts["moved"] = "any"
+		opts["padding"] = []int{0, 1, 0, 1}
+		opts["wrap"] = false
+		opts["close"] = "click"
+
+		g.ChannelCall("popup_create", params.Message, opts)
```
myitcv: We need to schedule the showing of a popup like this:

https://github.com/govim/govim/blob/40a414920272565dd9dcda9de5009a492858e66d/cmd/govim/gopls_client.go#L90

I also need to write up some docs/wiki explaining the threading model in `govim` and why this is necessary, but that's beyond this review.

cmd/govim/gopls_client.go
```diff
@@ -20,30 +20,47 @@ var _ protocol.Client = (*govimplugin)(nil)
 
 func (g *govimplugin) ShowMessage(ctxt context.Context, params *protocol.ShowMessageParams) error {
 	g.logGoplsClientf("Asked to show a message: %v", params.Message)
+	if params.Type == protocol.Error {
```
myitcv: In this situation, we only want to do something if the message is an error message.

So it's probably cleaner/easier to read if we return early when we don't have an error.

cmd/govim/test_config.go
```diff
@@ -75,6 +79,13 @@ func (v *vimstate) dumpPopups(args ...json.RawMessage) (interface{}, error) {
 	return sb.String(), nil
 }
 
+func (v *vimstate) showMessagePopup(args ...json.RawMessage) (interface{}, error) {
+	ctx := context.Background()
```
myitcv: Yep, this is exactly the kind of function we will need.

But what we need to do is asynchronously trigger the message in a go routine. 

You can return whatever you like, because the popup will be shown asynchronously.

cmd/govim/testdata/show_message.txt
```diff
@@ -0,0 +1,7 @@
+# Test that popup shows up when message is of Error type
+
+vim expr 'GOVIM_internal_ShowMessagePopup()'
+cmp stdout popup.golden
```
myitcv: We need to assert that a popup is actually shown; here we are just asserting the value returned by the function.

First, given the triggering of the `ShowMessage` is asynchronous, we need to wait until we see the call. This we do using `errlogmatch`:

https://github.com/govim/govim/blob/40a414920272565dd9dcda9de5009a492858e66d/cmd/govim/testdata/gofmt.txt#L10

You can instead wait using a pattern that will match the asynchronous `ShowMessage` callback (because you control the message that will appear in the logs).

Then you can "dump popups" to verify that the popup message shown is what you expect:

https://github.com/govim/govim/blob/40a414920272565dd9dcda9de5009a492858e66d/cmd/govim/testdata/function_hover.txt#L14

With any of this, if you want to turn on `govim` logging to stderr as you debug this test, take a look at:

https://github.com/govim/govim/wiki/govim-tests#debugging-test-failures

dragfire:
@myitcv I've pushed the changes as per the code review. Please check.

cmd/govim/testdata/show_message.txt
```diff
@@ -0,0 +1,19 @@
+# Test that popup shows up when message is of Error type
+
+[!vim] [!gvim] skip
+
+# For some reason, the new-style popup test fails on gvim.
+# https://github.com/govim/govim/issues/351
+[!vim:v8.1.1649] skip
```
myitcv: In https://github.com/govim/govim/pull/501 we increased the minimum version of Vim required to run `govim` to v8.1.1711, so conditions like this are now redundant.

I've just raised https://github.com/govim/govim/issues/518 to tidy this up everywhere else.

cmd/govim/testdata/show_message.txt
```diff
@@ -0,0 +1,19 @@
...
+
+vim expr 'GOVIM_internal_ShowMessagePopup()'
+cmp stdout showMessage.golden
+
+errlogmatch -wait 10s 'ShowMessage callback: Something went wrong' 
+[vim:v8.1.1649] vim -stringout expr 'GOVIM_internal_DumpPopups()'
```
myitcv: Same comment as above.

cmd/govim/testdata/show_message.txt
```diff
@@ -0,0 +1,19 @@
+# Test that popup shows up when message is of Error type
+
+[!vim] [!gvim] skip
+
+# For some reason, the new-style popup test fails on gvim.
+# https://github.com/govim/govim/issues/351
+[!vim:v8.1.1649] skip
+
+vim expr 'GOVIM_internal_ShowMessagePopup()'
+cmp stdout showMessage.golden
```
myitcv: This is really superfluous for this test. We have tests for function return values elsewhere

cmd/govim/gopls_client.go
```diff
@@ -19,31 +19,53 @@ const (
...
+		return nil
+	}
+
+	g.Schedule(func(govim.Govim) error {
+		opts := make(map[string]interface{})
+		opts["mousemoved"] = "any"
+		opts["moved"] = "any"
+		opts["padding"] = []int{0, 1, 0, 1}
+		opts["wrap"] = false
+		opts["close"] = "click"
```
myitcv: Just to confirm: this will place the popup in the middle of the screen, correct?

I think we probably want to be a bit less intrusive than that, add a bit of border and set a suitable highlight colour given the severity of the message. 

So how about the equivalent of:

```vim
call popup_create("we have a problem", {"mousemoved":"any", "moved":"any", "padding":[0,1,0,1], "wrap": "false", "close":"click", "border":[], "highlight": "ErrorMsg", "line":1})
```

That gives something like the following:

Screen Shot 2019-09-09 at 11 16 18

i.e.

* horizontal centre, place near the top
* error message highlighting
* `=` border

dragfire: @myitcv Thank you so much for a detailed review. Appreciate it. Yes you are right, I have changed my code to reflect what you suggested.

While I was testing the popup, I noticed that the test `show_message.txt` was passing when `vim` was actually throwing an error for using `"wrap": false`. It should be instead `"wrap": "false"`. I just thought it was weird that the test was passing. Please refer the screenshots below.
Screen Shot 2019-09-09 at 10 02 49 AM
Screen Shot 2019-09-09 at 10 02 33 AM

cmd/govim/gopls_client.go
```diff
@@ -19,31 +19,56 @@ const (
...
+	if params.Type != protocol.Error {
+		return nil
+	}
+
+	g.Schedule(func(govim.Govim) error {
+		opts := make(map[string]interface{})
+		opts["mousemoved"] = "any"
+		opts["moved"] = "any"
+		opts["padding"] = []int{0, 1, 0, 1}
+		opts["wrap"] = "false"
```
myitcv: This should be:

```
opts["wrap"] = false
```

What we send from Go will be JSON-marshaled. Then on the other end, Vim will translate from the JSON value to the appropriate Vim value (and fail it can't).

VimScript does not have true boolean values: see `:help TRUE` for more details. `v:true` and `v:false` do exist as convenience variables.

So my example in VimScript from earlier was actually incorrect, it should have read:

```vim
call popup_create("we have a problem", {"mousemoved":"any", "moved":"any", "padding":[0,1,0,1], "wrap": v:false, "close":"click", "border":[], "highlight": "ErrorMsg", "line":1})
```

cmd/govim/gopls_client.go
myitcv: See comment above.

myitcv:
Also, forgot to mention - please remove the `[WIP]` prefix before you next push in order to trigger builds for all entries in the build matrix.

dragfire:
@myitcv I was trying to rebase and messed up which resulted in my branch diverged. So I made another PR #519 instead of this. I hope you don't mind. Apologies for this.

cmd/govim/gopls_client.go
dragfire: @myitcv I have fixed this in #519 https://github.com/dragfire/govim/blob/show_message_popup/cmd/govim/gopls_client.go#L32

dragfire:
> @dragfire thank you; looks good pending one final change please.
> 
> Before we get this merged, please can you have a read through https://github.com/govim/govim/wiki/Contributors#pull-request-contributions. The commits in this PR will need to be squashed and a good commit message provided.
> 
> Incidentally, as you can probably tell we're in the early stages of welcoming new contributors, so any feedback on what does/doesn't work would be much appreciated.

@myitcv I struggled a bit with testing, it was mostly because I wasn't familiar with `testscript` and how certain things are being tested here. For example: `errlogmatch -wait 10s 'ShowMessage callback: Something went wrong' `. 

Overall I enjoyed working on this repo, got to learn a lot while implementing this small enhancement. You have been very helpful. Thank you for that.

myitcv:
@dragfire thanks for the feedback on the process. On thing I think I need to make clearer is how to get setup ready to do rebasing; that's easier configured right from the start.

I've actually fixed this PR with the changes from your other PR. With `git` it's actually very hard to _totally_ mess things up 😄  But in this situation things were definitely made more complicated by the fact that you had merged changes into your branch from `master`. I'll make sure I cover in the notes how to avoid this.

myitcv:
@dragfire for more information on `testscript` tests, please see https://github.com/govim/govim/wiki/govim-tests. Any suggestions on how that content can be improved, again, gratefully received.

Issue #507, opened 2019-09-07
cmd/govim: investigate gopls.Definition flake


myitcv:
```
--- FAIL: TestScripts (17.71s)
    --- FAIL: TestScripts/scripts (0.00s)
        --- FAIL: TestScripts/scripts/go_to_def_useopen (0.52s)
            testscript.go:308: 
                # Test that GOVIMGoToDef works
                # More natural to split below and to the right (0.027s)
                # Definition in an open window in current tab (0.252s)
                > vim ex 'set switchbuf=useopen'
                > vim ex 'e '$WORK/q/q.go
                > vim expr '[winnr(), tabpagenr()]'
                [stdout]
                [1,1]
                > stdout '^\Q[1,1]\E$'
                > vim ex 'split '$WORK/p.go
                > vim ex 'call cursor(5,17)'
                > vim ex 'GOVIMGoToDef'
                [stderr]
                unexpected command error: failed to ex(GOVIMGoToDef) in Vim: Caught 'got error whilst handling command:GOVIMGoToDef: failed to call gopls.Definition: no CheckPackageHandles for file://$WORK/q/q.go
                params were: &protocol.TextDocumentPositionParams{
                    TextDocument: protocol.TextDocumentIdentifier{URI:"file://$WORK/p.go"},
                    Position:     protocol.Position{Line:4, Character:16},
                }' in function <SNR>17_define[40]..<SNR>17_callbackCommand, line 5
                [exit status 255]
                FAIL: testdata/go_to_def_useopen.txt:14: unexpected command failure
```

Full build log https://gist.github.com/myitcv/5e2404143be2e2ff530ae8930129f29e

This is a `gopls` error which has been reported as https://github.com/golang/go/issues/34131

myitcv:
Marking as "onhold" because ultimately this is caused by `gopls` bug

myitcv:
Closing this. If the bug is fixed, great, otherwise we need to wait for #591

PR #509, opened 2019-09-07
cmd/govim: prevent non-batch calls being called when in batch

### LINKED ISSUE #438: cmd/govim: panic when trying to call non-batch methods in batch
We should panic if we attempt to call a non-batch channel method when in a batch.

This is a reminder to fix the code to do just that.

myitcv:
We should panic if we attempt to call a non-batch channel method when in a batch.

Fixes #438

### DIFF
--- a/cmd/govim/test_config.go
+++ b/cmd/govim/test_config.go
@@ -18,12 +18,13 @@ const (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


myitcv:
Merging (known flake in CI run #507)

PR #512, opened 2019-09-07
cmd/govim: completion options to disable fuzzy or deep matching

myitcv:
Provide options to disable the now-default fuzzy and/or deep completion
modes.

Whilst we await https://github.com/golang/go/issues/32258, users will
need to set these config options in their .vimrc and then restart Vim.
Even then there is a race condition for Vim8 package users that might
mean even this doesn't work.


myitcv:
> This might not be the place, but shouldn't we add tests for this as well?

Because we are waiting on https://github.com/golang/go/issues/32258, testing this because a little tricky: the config has to be in place before `gopls` starts. I've tried this out manually for now and all looks good. 

I've just raised https://github.com/govim/govim/issues/520 to remind us to add these tests.

leitzler:
:+1: I thought mostly about the static config tests, to see that the actual features fuzz & deep works. It would also detect things like "Ops, we accidentally changed the default to disable deep & fuzz" in gopls as a bonus (not like that would ever happen :)).

Hence the doubt, is it enough for us to test that the config is set and then assume that gopls works as intended?

myitcv:
Good point. We should also add those tests. I'll do that

myitcv:
Ok, I've just raised #521 in order to actually enable deep fuzzy by default 😄 (turns out I was running a special version locally)

Added a test for deep fuzzy in this PR.

Once #521 is merged we can look again at this PR

PR #515, opened 2019-09-08
cmd/govim: add info popup config to minimal.vimrc; add minimal.gvimrc

### LINKED ISSUE #392: cmd/govim: work out how to use previewpopup for completions
As of https://github.com/vim/vim/commit/7964873afe59d0896a921b7c585167674bb784d5, Vim can now use a preview popup to show information at complete time (assuming `completeopt` includes `preview`).

Investigate whether/how we should best leverage this, by perhaps putting docs in the preview popup?

myitcv:
Now that Vim supports popups for the info of a completion candidate, and
that popup window can be positioned relative to the candidates menu,
provide a version-guarded suggestion in the minimal.vimrc.

Also create a minimal.gvimrc so that Gvim users don't miss out on this
suggestions. This will require a tweak to the wiki in a couple of
places.

Fixes #392

### DIFF
--- a/cmd/govim/config/minimal.gvimrc
+++ b/cmd/govim/config/minimal.gvimrc
@@ -1,26 +1,7 @@
-" This file represents the minimal .vimrc needed to work with govim.
+" This file represents the minimal .gvimrc needed to work with govim.
 "
 " We also include a number of suggested settings that we think the majority of
 " users will like/prefer.
-
-set nocompatible
-set nobackup
-set nowritebackup
-set noswapfile
-
-filetype plugin on
-
-set mouse=a
-
-" To get hover working in the terminal we need to set ttymouse. See
-"
-" :help ttymouse
-"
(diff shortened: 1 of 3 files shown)


cmd/govim/config/minimal.gvimrc
```diff
@@ -0,0 +1,41 @@
...
+set updatetime=500
+
+" Suggestion: To make govim/Vim more responsive/IDE-like, we suggest a short
+" balloondelay
+set balloondelay=250
+
+" Suggestion: Turn on the sign column so you can see error marks on lines
+" where there are quickfix errors. Some users who already show line number
+" might prefer to instead have the signs shown in the number column; in which
+" case set signcolumn=number (requires Vim >= v8.1.1564)
```
leitzler: drop version comment as per #525

myitcv: Thanks - good catch.

PR #519, opened 2019-09-09
cmd/govim: show popup when MessageType is Error

### LINKED ISSUE #445: cmd/govim: properly handle LSP client ShowMessage

dragfire:
Existing implementation just logs the message. This implementation shows a popup when MessageType is error.

Fixes #445


myitcv:
We can close this because we'll fix things up in #508

PR #525, opened 2019-09-11
all: drop any references to now-old Vim versions

### LINKED ISSUE #518: all: remove references and conditions for now-insufficient Vim versions
In https://github.com/govim/govim/pull/501 we increased the minimum version of Vim required to run `govim`.

However, as part of this PR we didn't remove all references/conditions that refer to now-insufficient Vim versions, e.g. 

https://github.com/govim/govim/blob/230f08554a20eb7c4c4649ac87b358948bc489b6/cmd/govim/testdata/hover.txt#L7

We should complete this tidy up.

myitcv:
Following 63eec53 we now have a new minimum Vim version requirement for
govim: v8.1.1711. Therefore, any references to older versions are
defunct. Remove them, along with any conditions that were based on
previous versions.

As part of this we remove the now-legacy govim.ParseVimVersion to a test
file in testdriver because there doesn't appear to be a good way to get
the value of v:versionlong from the command line. Its accompanying test
moves with it.

We drop the condition test on the basis there's sufficient coverage by
virtue of conditions in testscript tests; no specific test is required.

Fixes #518


myitcv:
Previously discussed with @leitzler - merging.

PR #526, opened 2019-09-11
all: drop any references to now-old Vim versions

myitcv:
Follow up to 01d4e82; we missed dropping some condition references to
[!v8.1.1711] which is used to refer to versions prior to the now-minimum
version.

For #518


myitcv:
Will merge when CI passes as this is a simple follow up to #525

PR #529, opened 2019-09-12
cmd/govim: add config option for case sensitive completion

myitcv:


cmd/govim/config/config.go
```diff
@@ -60,6 +60,12 @@ type Config struct {
 	// Staticcheck enables staticcheck analyses in gopls
 	Staticcheck *bool `json:",omitempty"`
 
+	// CompletionCaseSensitive configures gopls' matching to be case sensitive.
+	// This option only takes effect when both CompletionDeepCompletionsDisable
```
leitzler: "when both CompleteDeepCompletions and CompletionFuzzyMatching are set to false..."

myitcv: Thanks. Fixed

PR #532, opened 2019-09-16
[WIP] quick hack to demo suggested fixes via popup

leitzler:
DO NOT MERGE & DO NOT REVIEW

Currently this requires a filter function defined outside govim. I added it initially to govim but it is called quite frequent so the logs will get spammed.

1) Add to `.vimrc`:
```
func SuggestedFixesFilter(id, key)                                                                                                                               
    if a:key == "n"                                                                                                                                              
        GOVIMSuggestedFixes                                                                                                                                      
        return 1                                                                                                                                                 
    endif                                                                                                                                                        
                                                                                                                                                                 
    return popup_filter_menu(a:id, a:key)                                                                                                                        
endfunc
```

2) Create a new main.go:
```
$ cd $(mktemp -d) && go mod init foo
$ cat > main.go
package main

func main() {
        x := 0
        y := 0
        x, y = x, y // will trigger "self assignment" warning with suggested edit
}
```

3) Open `main.go` in vim, and move cursor to line 6, run `GOVIMSuggestedFixes`
4) Press `n` to show next warning on that line, esc to abort and enter/space to perform fix


leitzler:
[asciicast](https://asciinema.org/a/taR70qcWa8TBDVATSZoM10sX1)

(note that the blinking block cursor is just an asciinema artifact)

myitcv:
This seems like a good starting point, with the only one suggestion being that we don't obscure the thing we're trying to fix 😄

leitzler:
I've moved the popup to cursor+1, as well as added support for multiple suggestions per diagnostic. I don't think there are any in gopls yet but as they arrive (with a newer version of staticcheck for example) they will work.

Things to decide on:
- If there are more than one diag on a line the user have to cycle between popups using 'n'. Is this OK UX? (it's kind of an edge case). Should it be 'n' and/or several others, like arrow right or 'l' (in line with hjkl-navigation)?
- When the user asks for suggested fixes it uses "current line" as baseline. We could, in theory, instead only show suggested fixes that are related to the specific cursor position. I wouldn't recommend it though since I have seen diagnostics that doesn't specify a.. "correct" range, but instead the same start & end position. It would be annoying to be required to have the cursor on that exact char.
- Tests using popup_getoptions (to fetch title) doesn't work at the moment (hence I haven't added the tests to this branch yet). I could go for "errlogmatch"-ing the popup_create call until that issue is resolved. It only affects testing.
- Multiple diags on the same line are sorted by title currently (just to make it consistent). Should it also sort on other things? I don't think the "preferred"-flag is used in gopls yet, but that might be a candidate. As well as severity obviously.

myitcv:
Please drop the `[WIP]` and ensure we are green across the board before merging.

PR #533, opened 2019-09-17
govim: use Vim's new SafeState* and state() to schedule work

myitcv:
As of e805441 we use a govim-local mechanism for scheduling work when it
is safe in Vim.

As of Vim v8.1.2056, we can leverage Vim-provided state/callbacks to
schedule this work, via the state() functions and SafeState*
autocommands.

In this change we leave the old govim-local mechanism in place for older
Vim versions, but switch to the state()/SafeState* approach for
sufficiently new versions.


myitcv:
@leitzler if you could give this a look over that would be great. If you could also try and run it locally to give it some more exposure that would be even better

myitcv:
Currently pending an answer to https://groups.google.com/d/msg/vim_dev/op_PKiE9iog/9wuOM78aAwAJ

PR #534, opened 2019-09-18
[WIP] cmd/govim: refactor diagnostics handling

leitzler:
To be able to add highlighting of diagnostics (over the entire range), diagnostics hovering, etc. we need to extract more things from the LSP diagnostics, like range and severity.

Much of the current logic take place inside the quickfix handling today, and the idea with this change is to move the generic stuff out from the quickfix specific code. That includes sign placement as well, but we still want to ensure a consistent state where signs always represents what is shown in the quickfix list.

### DIFF
--- a/cmd/govim/vimstate.go
+++ b/cmd/govim/vimstate.go
@@ -100,7 +100,7 @@ func (v *vimstate) setConfig(args ...json.RawMessage) (interface{}, error) {
 			v.diagnosticsLock.Lock()
 			v.diagnosticsChanged = true
 			v.diagnosticsLock.Unlock()
-			return nil, v.updateQuickfix()
+			return nil, v.redefineDiagnostics()
 		}
 	}
 	return nil, nil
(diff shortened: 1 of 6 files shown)


cmd/govim/vimstate.go
```diff
@@ -113,7 +118,95 @@ func (v *vimstate) setUserBusy(args ...json.RawMessage) (interface{}, error) {
 	if v.userBusy || v.config.QuickfixAutoDiagnosticsDisable || !v.quickfixIsDiagnostics {
 		return nil, nil
 	}
-	return nil, v.updateQuickfix()
+	return nil, v.redefineDiagnostics()
+}
+
+func (v *vimstate) redefineDiagnostics() error {
```
myitcv: I'd suggest perhaps creating `diagnostics.go` to split this file up a bit? It's become a bit of a grab bag

leitzler: Good point, fixed

PR #542, opened 2019-09-23
cmd/govim: add config to enable staticcheck within gopls

myitcv:
Disable staticcheck by default for now whilst we iron out inevitable
issues.

Add a test to verify it works.


myitcv:
Currently blocked on https://github.com/golang/go/issues/34494

myitcv:
Agreed. Added a comment about tracking https://github.com/golang/go/issues/34494 to the test where we have the "workaround" - when https://github.com/golang/go/issues/34494 is fixed, the test will break in any case.

PR #543, opened 2019-09-23
cmd/govim: switch to better testscript structure

myitcv:
cmd/govim: switch to better testscript structure

Move all cmd/govim test scripts into a directory structure under
testdata:

* testdata/install is for long test install-time tests that verify the
installation and use of govim via the advertised plugin mechanisms
* testdata/scenario_default contains scripts that verify the default
configuration of govim
* testdata/scenario_* contains other default configurations, where each
directory can contain a default_config.json file that defines a
cmd/govim/config.Config to use as the defaults

Because we are no longer guaranteed of a single set of defaults, we have
to be more defensive when using cmd/govim/config.Config values. Hence,
we make relevant changes in cmd/govim, defining a convenience function
in cmd/govim/internal/vimconfig to help with that.

Add testdata/scenario_caseinsensitivecompletion to verify case
insensitive completions, where deep and fuzzy is turned off.


cmd/govim/internal/vimconfig/vimconfig.go
```diff
@@ -66,3 +66,16 @@ func BoolVal(v bool) *bool {
 func MapVal(v map[string]interface{}) *map[string]interface{} {
 	return &v
 }
+
+// EqualBool returns true iff i and j are both nil, or if both are non-nil and
```
leitzler: typo - `iff`

leitzler: incorrect comment, "returns true if i and j are both nil" and then `return false` if both are nil.

cmd/govim/main_test.go
```diff
@@ -50,75 +52,128 @@ func TestScripts(t *testing.T) {
...
+							if err := json.Unmarshal(config, &defaults); err != nil {
+								t.Fatalf("failed to unmarshal JSON config in %v: %v\n%s", configPath, err, config)
+							}
+							// Now verify that we haven't supplied superfluous JSON
+							var i interface{}
+							if err := json.Unmarshal(config, &i); err != nil {
+								t.Fatalf("failed to re-parse JSON config in %v: %v\n%s", configPath, err, config)
+							}
+							var defaultsCheck []byte
+							{
```
leitzler: What is the purpose of this additional scope?

cmd/govim/internal/vimconfig/vimconfig.go
myitcv: I intentionally used `iff` there: https://en.wikipedia.org/wiki/If_and_only_if

Thanks - the code was actually wrong here 😢 - fixed.

cmd/govim/main_test.go
myitcv: Nothing more than an artefact of a refactoring. Fixed.

PR #548, opened 2019-09-26
deps: upgrade to gopls and x/tools a8d5d342

myitcv:
* internal/lsp: provide option for case sensitive completion a8d5d342


myitcv:
Merging because this doesn't require any `govim` changes

PR #551, opened 2019-09-28
cmg/govim: add env var to pass flags to gopls

leitzler:
This change adds a new environment variable, GOVIM_GOPLS_FLAGS, that is used to pass additional flags to gopls.

### DIFF
--- a/cmd/govim/internal/util/util.go
+++ b/cmd/govim/internal/util/util.go
@@ -1,0 +1,61 @@
+package util
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// Split breaks the line into words, evaluating quoted
+// strings and evaluating environment variables.
+// Implementation is a slightly modified version of go generate's
+// split(), from src/cmd/go/internal/generate/generate.go in stdlib
+func Split(line string) ([]string, error) {
+	// Parse line, obeying quoted strings.
+	var words []string
+	// There may still be a carriage return.
+	if len(line) > 0 && line[len(line)-1] == '\r' {
+		line = line[:len(line)-1]
+	}
+	// One (possibly quoted) word per iteration.
+Words:
+	for {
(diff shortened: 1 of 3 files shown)


cmd/govim/main.go
```diff
@@ -246,7 +246,12 @@ func (g *govimplugin) Init(gg govim.Govim, errCh chan error) error {
 
 	g.ChannelExf("let s:gopls_logfile=%q", logfile.Name())
 
-	gopls := exec.Command(g.goplspath, "-rpc.trace", "-logfile", logfile.Name())
+	goplsArgs := []string{"-rpc.trace", "-logfile", logfile.Name()}
+	if flags := strings.Fields(os.Getenv(string(config.EnvVarGoplsFlags))); flags != nil {
```
myitcv: I think we should look to handle quoted strings for values that can include spaces. Not because I have an immediate use case for it, but because a) people will expect it and b) there's an implementation we can use.

Take a look at https://github.com/golang/go/blob/af880809d8c3896a7d37d7547c290cd9bc34530f/src/cmd/go/internal/generate/generate.go#L332

leitzler: Added

cmd/govim/internal/util/util.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		i := strings.IndexAny(line, " \t")
+		if i < 0 {
+			i = len(line)
+		}
+		words = append(words, line[0:i])
+		line = line[i:]
+	}
+	// Substitute environment variables.
+	for i, word := range words {
```
myitcv: We don't want to expand in this case, because values will already have been expanded in the setting of `GOVIM_GOPLS_FLAGS`. Expansion is required in `go generate` because we are evaluating literals in a `.go` file. So this loop can be dropped

leitzler: fixed

PR #564, opened 2019-11-02
cmd/govim: add highlights for all LSP severities and use them for signs

leitzler:
This change adds default highlights for the different LSP severities.
They can all be overridden in your .vimrc to match your theme. If you
for example would prefer blue background for errors in your color
terminal you'd add this line to your .vimrc:

hi govimErr ctermbg=Blue

The different highlight names are:
- govimErr
- govimWarn
- govimInfo
- govimHint

This is also a gardening change for diagnostics highlights that will
use the same defined highlights.

### DIFF
--- a/cmd/govim/internal/types/types.go
+++ b/cmd/govim/internal/types/types.go
@@ -222,5 +222,23 @@ type Diagnostic struct {
 	Range    Range
 	Text     string
 	Buf      int
-	Severity int
+	Severity Severity
 }
+
+// Severity is the govim internal representation of the LSP DiagnosticSeverites
+type Severity int
+
+const (
+	SeverityErr  = Severity(protocol.SeverityError)
+	SeverityWarn = Severity(protocol.SeverityWarning)
+	SeverityInfo = Severity(protocol.SeverityInformation)
+	SeverityHint = Severity(protocol.SeverityHint)
+)
+
+// Highlight names as defined by ftplugin/go.vim.
+const (
+	HighlightErr  string = "govimErr"
+	HighlightWarn string = "govimWarn"
+	HighlightInfo string = "govimInfo"
+	HighlightHint string = "govimHint"
+)
--- a/cmd/govim/signs.go
+++ b/cmd/govim/signs.go
@@ -9,10 +9,27 @@ import (
 // Using a sign group creates a separate namespace for all signs placed by govim
 const signGroup = "govim"
 
-// Name of different sign types used, only one at the moment, errors.
-const (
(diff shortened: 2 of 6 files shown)


cmd/govim/internal/types/types.go
```diff
@@ -222,5 +222,23 @@ type Diagnostic struct {
 	Range    Range
 	Text     string
 	Buf      int
-	Severity int
+	Severity Severity
 }
+
+// Severity is the govim internal representation of the LSP DiagnosticSeverites
+type Severity int
```
myitcv: Just to check: we're defining a new type here to deal with the fact that the protocol type is float?

ftplugin/go.vim
```diff
@@ -1,4 +1,10 @@
 if GOVIMPluginStatus() == "initcomplete"
+  " Highlights
+  highlight link govimErr Error
```
myitcv: Per our Slack conversation, this makes me a tiny bit nervous... because I don't think we can easily cater for everyone's respective themes. Hence whatever colours we choose here aren't ever going to be a great default for everyone. 

As we discussed however, the reason we need these values is:

* customisation in the case that folks want specific config for Go related errors/warnings
* the Vim highlight space doesn't have the option space, specifically it doesn't have definitions for warnings, info, hint etc

cmd/govim/signs.go
```diff
@@ -9,26 +9,45 @@ import (
...
 
-// Name of different sign types used, only one at the moment, errors.
-const (
-	errorSign = "govimerr"
-)
+// signSuffix is used when naming the different sign types. It is appended to each
+// highlight type and passed in as name to sign_define().
+const signSuffix = "Sign"
+
+// Each sign in vim has a prio. If there are multiple signs on the same line, it
```
myitcv: I'm assuming "prio" is "priority"? If so, perhaps we can rename this - it's not a shortening I'm particularly familiar with and it doesn't seem worth the potential confusion for the sake of 4 chars 😄 (admittedly everywhere)

cmd/govim/internal/types/types.go
leitzler: Yes, that floating severity in the protocol is.. yeah.

ftplugin/go.vim
leitzler: Yeah, I agree. Let's monitor the feedback and see what users think about it. It is the best tradeoff I could come up with as a start.

cmd/govim/signs.go
leitzler: fixed

PR #566, opened 2019-11-03
[WIP] plugin: try to fix SafeState "race"

myitcv:


myitcv:
We do not believe this is the "right" way to fix the problem... or indeed that there is a problem :)

PR #573, opened 2019-11-06
cmd/govim: keep track of if buffers are "Loaded" or not

leitzler:
To be able to use textprops for highlighting, we must be aware of
if a buffer is loaded or not. Text properties can't be added to
unloaded buffers, and trying to do so will throw an "invalid line"
error.

This change keeps track of if buffers are loaded or not.

### DIFF
--- a/cmd/govim/internal/types/types.go
+++ b/cmd/govim/internal/types/types.go
@@ -23,6 +23,9 @@ type Buffer struct {
 	Version  int
 	Listener int
 
+	// Loaded reflects vim's "loaded" buffer state. See :help bufloaded() for details.
+	Loaded bool
+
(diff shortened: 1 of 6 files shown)


cmd/govim/internal/types/types.go
```diff
@@ -23,6 +23,9 @@ type Buffer struct {
 	Version  int
 	Listener int
 
+	// Loaded reflects vim's "loaded" flag. See :help bufloaded() for details.
```
myitcv: What does "flag" mean here? 

I think `loaded` represents part of Vim's state with respect to a buffer.

leitzler: updated

PR #579, opened 2019-11-10
cmd/govim: text propoperty highlight diagnostics

leitzler:
Diagnostics from gopls will now be highlighted directly in the code. There are four different highlights that can be overridden in vim to use different styles/colors for each of the LSP severities Error, Warning, Info and Hint.

Note that this change also renames sign highlights, which is a breaking change for anyone that use custom colors for their signs.

[asciicast](https://asciinema.org/a/33WJNxjZs1vdZQYWv4DwVy6qr)

### DIFF
--- a/cmd/govim/buffer_events.go
+++ b/cmd/govim/buffer_events.go
@@ -29,6 +29,7 @@ func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
 	if cb, ok := v.buffers[nb.Num]; ok {
 		// reload of buffer, e.v. e!
 		cb.SetContents(nb.Contents())
+		cb.Loaded = nb.Loaded
 		cb.Version++
 		return v.handleBufferEvent(cb)
 	}
@@ -130,6 +131,10 @@ func (v *vimstate) bufUnload(args ...json.RawMessage) error {
 }
 
 func (v *vimstate) handleBufferEvent(b *types.Buffer) error {
+	if err := v.redefineHighlights(v.diagnostics()); err != nil {
+		v.Logf("failed to update highlights for buffer %d: %v", b.Num, err)
+	}
+
 	v.triggerBufferASTUpdate(b)
 	if b.Version == 1 {
 		params := &protocol.DidOpenTextDocumentParams{
--- a/cmd/govim/config/config.go
+++ b/cmd/govim/config/config.go
@@ -50,6 +50,14 @@ type Config struct {
 	// Default: true
 	QuickfixSigns *bool `json:",omitempty"`
 
+	// HighlightDiagnostics enables in-code highlighting of diagnostics using
+	// text properties. Each diagnostic reported by gopls will be highlighted
+	// according to it's severity, using the following vim defined highlight
+	// groups: GOVIMErr, GOVIMWarn, GOVIMInfo & GOVIMHint.
+	//
+	// Default: true
+	HighlightDiagnostics *bool `json:",omitempty"`
+
 	// CompletionDeepCompletiions enables gopls' deep completion option
 	// in the derivation of completion candidates.
 	//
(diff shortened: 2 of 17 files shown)


cmd/govim/highlight.go
```diff
@@ -0,0 +1,80 @@
...
+		if v.ParseInt(res[i]) != 0 {
+			return fmt.Errorf("call to prop_type_add() failed")
+		}
+	}
+	return nil
+}
+
+func (v *vimstate) redefineHighlights(diags []types.Diagnostic) error {
+	v.BatchStart()
+	defer v.BatchEnd()
```
myitcv: You want `v.BatchCancelIfNotEnded()` in this defer and should instead explicitly call `v.BatchEnd()` on the happy path below

testsetup/testsetup.go
```diff
@@ -25,7 +25,8 @@ const (
 
 // user environment variables
 const (
-	EnvLogfileTmpl = "GOVIM_LOGFILE_TMPL"
+	EnvLogfileTmpl           = "GOVIM_LOGFILE_TMPL"
+	EnvDisableHighlightDiags = "GOVIM_DISABLE_HIGHLIGHTDIAGS"
```
myitcv: I don't think we need the environment variable master override in this case.

Instead we should just add a regular `Config` value to disable this:

https://pkg.go.dev/github.com/govim/govim/cmd/govim/config?tab=doc#Config

cmd/govim/main.go
```diff
@@ -428,3 +435,29 @@ func (g *govimplugin) Shutdown() error {
 	}
 	return nil
 }
+
+func (g *govimplugin) defineHighlights() {
+	warnColor := 166 // Orange
+	if vimColors, err := strconv.Atoi(g.ParseString(g.ChannelExpr(`&t_Co`))); err != nil || vimColors < 256 {
+		warnColor = 3 // Yellow, fallback when the terminal doesn't support at least 256 colors
+	}
+	g.ChannelExf("highlight default %s term=underline cterm=underline gui=undercurl ctermfg=1 guisp=Red", types.HighlightErr)
```
myitcv: Batch these calls? Not critical if we don't, but will help to reduce startup time by a few ms.

cmd/govim/main.go
```diff
@@ -428,3 +435,29 @@ func (g *govimplugin) Shutdown() error {
...
+	g.ChannelExf("highlight default %s ctermfg=15 ctermbg=%d guisp=Orange guifg=Orange", types.HighlightSignWarn, warnColor)
+	g.ChannelExf("highlight default %s ctermfg=15 ctermbg=6 guisp=Cyan guifg=Cyan", types.HighlightSignInfo)
+	g.ChannelExf("highlight default link %s %s", types.HighlightSignHint, types.HighlightSignInfo)
+}
+
+func (g *govimplugin) highlightDiagnostics() bool {
+	if g.Flavor() != govim.FlavorVim && g.Flavor() != govim.FlavorGvim {
+		return false
+	}
+	if os.Getenv(testsetup.EnvDisableHighlightDiags) == "true" {
```
myitcv: With a normal config value this method can, I think, go?

myitcv: You will however need to add to the logic in

```
func (v *vimstate) setConfig(args ...json.RawMessage) (interface{}, error)
```

testsetup/testsetup.go
leitzler: fixed

cmd/govim/main.go
leitzler: indeed, removed and added config instead

cmd/govim/buffer_events.go
```diff
@@ -130,6 +131,10 @@ func (v *vimstate) bufUnload(args ...json.RawMessage) error {
 }
 
 func (v *vimstate) handleBufferEvent(b *types.Buffer) error {
+	if err := v.redefineHighlights(v.diagnostics()); err != nil {
+		v.Logf("failed to update highlights for buffer %d: %v", b.Num, err)
```
myitcv: In a later PR I think we perhaps need to show the user an error popup in addition to logging.

Which means it might well be time therefore to extract something common from:

https://github.com/govim/govim/blob/3376e910e567b9a0e09994f0f0f5cb6c82f3f5f8/cmd/govim/gopls_client.go#L28-L57

cmd/govim/config/config.go
```diff
@@ -50,6 +50,14 @@ type Config struct {
 	// Default: true
 	QuickfixSigns *bool `json:",omitempty"`
 
+	// HighlightDiagnostics enables in-code highlighting of diagnostics using
+	// text properties. Each diagnostic reported by gopls will be highlighted
+	// according to it's severity, using the following vim defined highlight
+	// rules: govimErr, govimWarn, govimInfo & govimHint.
```
myitcv: Are these "rules"?

leitzler: No, docs refer to them as "groups", will fix.

cmd/govim/buffer_events.go
leitzler: Indeed, wrote #641 as reminder

cmd/govim/config/config.go
```diff
@@ -50,6 +50,14 @@ type Config struct {
 	// Default: true
 	QuickfixSigns *bool `json:",omitempty"`
 
+	// HighlightDiagnostics enables in-code highlighting of diagnostics using
+	// text properties. Each diagnostic reported by gopls will be highlighted
+	// according to it's severity, using the following vim defined highlight
```
cespare: its

PR #580, opened 2019-11-12
deps: upgrade to gopls and x/tools a3f652f1

myitcv:


myitcv:
FYI I snuck in a check later version of these deps. Still green, still no changes required on the `govim` side

PR #590, opened 2019-11-21
cmd/govim: place signs for existing diagnostics in new buffers

leitzler:
If gopls report diagnostics for files that aren't opened in vim
they won't get any signs placed when they are opened.

A typical use case is when staticcheck report diagnostics for
a file that isn't opened, and the user uses quickfix to jump
directly to the issue (effectively opening a new buffer).

This change updates the internal diagnostics cache when a buffer
is read or closed so existing diagnostics will have signs placed.

### DIFF
--- a/cmd/govim/buffer_events.go
+++ b/cmd/govim/buffer_events.go
@@ -17,6 +17,15 @@ import (
 
 func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
 	nb := v.currentBufferInfo(args[0])
+
+	// If we load a buffer that already had diagnostics reported by gopls, the buffer number must be
+	// updated to ensure that sign placement etc. works.
+	for i, d := range v.diagnosticsCache {
+		if d.Buf == -1 && d.Filename == nb.URI().Filename() {
+			v.diagnosticsCache[i].Buf = nb.Num
+		}
+	}
+
 	if cb, ok := v.buffers[nb.Num]; ok {
 		// reload of buffer, e.v. e!
 		cb.SetContents(nb.Contents())
@@ -36,6 +45,13 @@ func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
 		nb.Version = 1
 	}
 	nb.Listener = v.ParseInt(v.ChannelCall("listener_add", v.Prefix()+string(config.FunctionEnrichDelta), nb.Num))
+
+	if v.placeSigns() {
+		if err := v.redefineSigns(v.diagnostics()); err != nil {
+			v.Logf("failed to update signs for buffer %d: %v", nb.Num, err)
+		}
+	}
+
 	return v.handleBufferEvent(nb)
 }
 
@@ -151,6 +167,17 @@ func (v *vimstate) deleteCurrentBuffer(args ...json.RawMessage) error {
 	if !ok {
 		return fmt.Errorf("tried to remove buffer %v; but we have no record of it", currBufNr)
 	}
+
+	// The diagnosticsCache is updated with -1 (unknown buffer) as bufnr.
+	// We don't want to remove the entries completely here since we want to show them in
+	// the quickfix window. And we don't need to remove existing signs or text properties
+	// either here since they are removed by vim automatically when a buffer is deleted.
+	for i := range v.diagnosticsCache {
+		if v.diagnosticsCache[i].Buf == currBufNr {
+			v.diagnosticsCache[i].Buf = -1
+		}
+	}
+
 	v.ChannelCall("listener_remove", cb.Listener)
 	delete(v.buffers, cb.Num)
 	params := &protocol.DidCloseTextDocumentParams{
--- a/cmd/govim/diagnostics.go
+++ b/cmd/govim/diagnostics.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"io/ioutil"
-	"path/filepath"
 	"sort"
 	"strings"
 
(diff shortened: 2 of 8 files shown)


leitzler:
I moved the new test to a separate file, since it seems like changing the entire buffer often trigger #591 (previously it was a part of signs.txt but with a separate main.go).

cmd/govim/buffer_events.go
```diff
@@ -17,6 +18,19 @@ import (
...
 	nb := v.currentBufferInfo(args[0])
+	rel, err := filepath.Rel(v.workingDirectory, nb.URI().Filename())
+	if err != nil {
+		v.Logf("bufReadPost: failed to call filepath.Rel(%q, %q): %v", v.workingDirectory, nb.URI().Filename(), err)
+	}
+
+	// If we load a buffer that already had diagnostics reported by gopls, the buffer number must be
+	// updated to ensure that sign placement etc. works.
+	for i, d := range v.diagnosticsCache {
+		if d.Buf == -1 && d.Filename == rel {
```
myitcv: I think the identification of the buffer should be on the full path, not the relative path. That way we make this part of the code at least agnostics of the cwd.

By all means calculate and cache the current rel path during lazy evaluation as a separate field

cmd/govim/buffer_events.go
```diff
@@ -17,6 +18,19 @@ import (
...
 func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
 	nb := v.currentBufferInfo(args[0])
+	rel, err := filepath.Rel(v.workingDirectory, nb.URI().Filename())
+	if err != nil {
+		v.Logf("bufReadPost: failed to call filepath.Rel(%q, %q): %v", v.workingDirectory, nb.URI().Filename(), err)
+	}
+
+	// If we load a buffer that already had diagnostics reported by gopls, the buffer number must be
+	// updated to ensure that sign placement etc. works.
+	for i, d := range v.diagnosticsCache {
```
myitcv: On reflection, possibly slightly more effective (both in terms of performance but more sort readability) for `diagnosticsCache` to be `map[string]*types.Diagnostic`?

The key can then be the absolute filepath.

cmd/govim/buffer_events.go
```diff
@@ -151,6 +172,17 @@ func (v *vimstate) deleteCurrentBuffer(args ...json.RawMessage) error {
 	if !ok {
 		return fmt.Errorf("tried to remove buffer %v; but we have no record of it", currBufNr)
 	}
+
+	// The diagnosticsCache is updated with -1 (unknown buffer) as bufnr.
+	// We don't want to remove the entries completely here since we want to show them in
+	// the quickfix window. And we don't need to remove existing signs or text properties
+	// either here since they are removed by vim automatically when a buffer is deleted.
+	for i := range v.diagnosticsCache {
```
myitcv: This then becomes a lookup by absolute filepath of the buffer (which we have in hand)

cmd/govim/diagnostics.go
```diff
@@ -52,9 +49,9 @@ func (v *vimstate) diagnostics() []types.Diagnostic {
 			buf = types.NewBuffer(-1, fn, byts, false)
 		}
 		// make fn relative for reporting purposes
-		fn, err := filepath.Rel(cwd, fn)
+		fn, err := filepath.Rel(v.workingDirectory, fn)
```
myitcv: Per previous comments, worth caching the relative filepath (and we can fix that being "wrong" later, per our Slack chat) but definitely worth making identity based on absolute filepath.

cmd/govim/main.go
```diff
@@ -312,9 +313,8 @@ func (g *govimplugin) Init(gg govim.Govim, errCh chan error) error {
 		g: g,
 	}
 
-	wd := g.ParseString(g.ChannelCall("getcwd", -1))
 	initParams := &protocol.ParamInitia{}
-	initParams.RootURI = string(span.FileURI(wd))
+	initParams.RootURI = string(span.FileURI(g.vimstate.workingDirectory))
```
myitcv: Agree, this is much less fragile for now. At least if it's wrong it will be consistently wrong from the point at which the user calls `cwd` onwards.

cmd/govim/buffer_events.go
leitzler: Indeed, I think we even talked about it but it must have slipped my mind, apologies.

cmd/govim/buffer_events.go
leitzler: Yes (I guess you mean `map[string][]*types.Diagnostic`). I'd still need to sort them in a slice for the quick fix list, but I wouldn't need to iterate over all diagnostics when placing signs & highlighting diagnostics.

cmd/govim/buffer_events.go
leitzler: Good point

cmd/govim/diagnostics.go
leitzler: Ack

cmd/govim/main.go
leitzler: 👍

cmd/govim/buffer_events.go
leitzler: (I recall that a nice side effect of sorting all diagnostics before placing signs as well was that I automatically got a consistent golden file in the tests.)

leitzler: Just had another quick look at this and indeed, both the sign placement and quickfix list needs to be sorted. Or rather "needs" to be sorted as signs only needs to be sorted for them to be testable with golden files, since they rely on the sign id's (placement order) to be consistent.

So from the three options:
- Change to map & convert to sorted slice sort twice (quickfix & signs)
- Change to map & convert once & rewrite sign-related tests so they don't rely on sign id:s
- Leave as is

I lean towards leaving it as is for now. That change could be a separate PR for the future. Thoughts?

myitcv: Yep, agreed.

myitcv: Leave as is for now

PR #593, opened 2019-11-25
cmd/govim: block govim initialisation until post Configuration call

myitcv:


cmd/govim/main.go
```diff
@@ -180,6 +180,9 @@ type govimplugin struct {
 	diagnosticsCache []types.Diagnostic
 
 	bufferUpdates chan *bufferUpdate
+
```
leitzler: Maybe add a comment with reference to "top of (*govimplugin.Configuration)" here as well so that we won't miss removing this as well when reverting this change.

myitcv: Done

PR #606, opened 2019-12-02
cmd/govim: add concept of user config in test overrides

myitcv:


cmd/govim/config/internal/applygen/main.go
```diff
@@ -0,0 +1,105 @@
...
+			fields[i] = st.Field(i).Name()
+		}
+		toGen[name] = fields
+	}
+	sort.Strings(names)
+
+	var buf bytes.Buffer
+
+	pf := func(format string, args ...interface{}) {
+		if format[len(format)-1] != '\n' {
```
leitzler: Any reason you aren't doing `format = strings.TrimSuffix(format, "\n") + "\n"` instead? (to avoid out of index panics, not that there is a risk for that here).

cmd/govim/config/internal/applygen/exitcode_plan9.go
```diff
@@ -0,0 +1,16 @@
+package main
```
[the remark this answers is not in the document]
myitcv: Simply that anything `syscall` related must be guarded by build tags, c.f. [Rob Pike](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=11m10s)

cmd/govim/config/internal/applygen/main.go
myitcv: This is just a simply crutch function here that avoids a bit of typing, not looking to be super advanced/useful. Keeps things a bit prettier later on. I could equally have done what you propose :)

PR #611, opened 2019-12-06
deps: use cherry pick of CL 210077 for memoize race fix

myitcv:
In our fork of tools at github.com/myitcvforks/tools we also cherry-pick
the commit from CL 210077 which contains a fix for the data races (and
hence CI failures) we reported as part of:

https://github.com/golang/go/issues/35995


myitcv:
Merging because this isn't particularly contentious. No changes in `govim` itself as part of this

PR #614, opened 2019-12-08
tests: define a default value for errlogmatch -wait flag

### LINKED ISSUE #607: tests: make errlogmatch -wait default be environment variable defined
In all of our testscript tests where we use `errlogmatch` we specify a `-wait` flag, e.g.

```
errlogmatch -wait 30s 'sendJSONMsg: .*\"call\",\"s:batchCall\",.*\"sign_placelist\"'
```

The figure of `30s` is an arbitrary figure deemed long enough to allow for slow running tests on CI (typically slower than dev machines) when running with `-race`.

This makes `30s` far too long however when running tests locally. 

It's also totally crazy to have to specify this flag everywhere - we should set a default value.

Combine these two observations to define a default value of `-wait` via an environment variable such that we can have tests fail quicker locally vs CI and obviate the need the always specify this flag.

myitcv:
errlogmatch's -wait flag is now optional. It's value is taken from
os.Getenv("GOVIM_ERRLOGMATCH_WAIT") else it is 30s, the current value.

Closes #607

### DIFF
--- a/testdriver/testdriver.go
+++ b/testdriver/testdriver.go
@@ -796,6 +796,11 @@ func ErrLogMatch(ts *testscript.TestScript, neg bool, args []string) {
 		ts.Fatalf("errlogmatch %v was not the right type", KeyErrLog)
 	}
 
+	defaultWait := os.Getenv("GOVIM_ERRLOGMATCH_WAIT")
+	if defaultWait == "" {
+		defaultWait = "30s"
+	}
+
 	fs := flag.NewFlagSet("errlogmatch", flag.ContinueOnError)
 	fStart := fs.Bool("start", false, "search from beginning, not last snapshot")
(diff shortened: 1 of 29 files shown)


testdriver/testdriver.go
```diff
@@ -805,8 +810,21 @@ func ErrLogMatch(ts *testscript.TestScript, neg bool, args []string) {
...
 
-	if *fWait != "" && neg {
-		ts.Fatalf("-wait is not compatible with negating the command")
+	var waitSet bool
+	fs.Visit(func(f *flag.Flag) {
+		if f.Name == "wait" {
+			waitSet = true
+		}
+	})
+
```
leitzler: Really nits, would this be more readable?
```
if waitSet && neg {
        ts.Fatalf(...)
}

if *fWait == "" && !neg {
        fWait = &defaultWait
}
```

myitcv: It is, thanks.

PR #617, opened 2019-12-09
ci: go1.13.5, go1.12.14 and latest Vim v8.1.2414

myitcv:


myitcv:
Nothing controversial here; will merge on successful build.

PR #619, opened 2019-12-10
[DO NOT MERGE] extra debug spam

heschik:


myitcv:
I think we can close this now that we have made progress in https://github.com/golang/go/issues/35638#issuecomment-565718553

PR #620, opened 2019-12-11
Detect go.mod as a 'gomod' filetype; add syntax and indent

cespare:
First, add ftdetect configuration to detect go.mod files as the 'gomod' filetype.

Then add syntax and indent files for this filetype.

The gomod syntax file is reused from the vim-go project under the terms of its 3-clause BSD license. The indent file is new work.

### DIFF
--- a/ftdetect/go.vim
+++ b/ftdetect/go.vim
@@ -1,0 +1,6 @@
+" For now, rely on the .go detection, syntax, indent, etc. provided in recent
+" Vim distributions.
+
+" By default, vim associates .mod files with filetypes lprolog or modsim3.
+" Override these rather than using setfiletype.
+autocmd BufNewFile,BufRead go.mod setlocal filetype=gomod
--- a/indent/gomod.vim
+++ b/indent/gomod.vim
@@ -1,0 +1,33 @@
+if exists("b:did_indent")
+  finish
+endif
+let b:did_indent = 1
+
+setlocal indentexpr=GomodIndent(v:lnum)
+
+if exists("*GomodIndent")
+  finish
+endif
+
+function! GomodIndent(lnum)
+  let prevnum = prevnonblank(a:lnum-1)
+  if prevnum == 0
+    return 0
+  endif
+
+  " Take the current and previous lines, trimming end-of-line spaces
+  " and comments.
+  let line = substitute(getline(a:lnum), '\v\s+(//.*)?$', '', '')
+  let prev = substitute(getline(prevnum), '\v\s+(//.*)?$', '', '')
+  let ind = indent(prevnum)
+
+  if prev =~ '\v^(require|replace|exclude).*\($'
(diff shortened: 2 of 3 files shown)


myitcv:
Thanks. Did you intend for this to also do something with respect to auto indenting in `go.mod` files?

For example, starting with:

```
module mod.com

go 1.14

require (_)
```

(which is typically a position I find myself in because I use [`auto-pairs`](https://github.com/jiangmiao/auto-pairs)) with the cursor at the position `_`, if I then hit `<Enter>` I get:

```
module mod.com

go 1.14

require (

    )
```

cespare:
Sure, I can write an indent file. I agree it would be nice to have. Looks like vim-go doesn't have one, though.

cespare:
@myitcv I wrote a simple indent file for go.mod. I've never written an indent file before but this seems like a pretty basic filetype so hopefully this will cover all the important cases for now. PTAL.

ftdetect/go.vim
```diff
@@ -0,0 +1,6 @@
+" For now, rely on the .go detection, syntax, indent, etc. provided in recent
+" Vim distributions.
+
+" By default, vim associates .mod files with filetypes lprolog or modsim3.
+" Override these rather than using setfiletype.
+autocmd BufNewFile,BufRead go.mod setlocal filetype=gomod
```
myitcv: This doesn't feel quite right here. Is there a reason it needs to be here, or could it live in `plugin/govim.vim`?

cespare: AFAIK putting this in the `ftdetect` directory is the standard layout for Vim plugins. It also has special semantics: vim automatically creates an augroup for the commands in the ftdetect file.

cespare:
> Then if you don't mind squashing the commits and providing a helpful commit message in the mould of previous commits https://github.com/govim/govim/commits/master. Use a prefix of `cmd/govim` to indicate this relates to the plugin.

Done, though it seems like a strange convention to use `cmd/govim` to tag changes which are entirely distinct from the cmd/govim code and which do not touch the cmd/govim path.

myitcv:
> though it seems like a strange convention to use cmd/govim to tag changes which are entirely distinct from the cmd/govim code and which do not touch the cmd/govim path.

There is some attempt at logic behind this rather strange approach. 

`github.com/govim/govim` is a package in and of itself that should be used by people looking to write Vim plugins in Go.

`github.com/govim/govim/cmd/govim` is the plugin commonly known as `govim`. 

Using the Go project convention here of using package paths for issue/PR prefixes, because this change relates to `github.com/govim/govim/cmd/govim` (albeit Vim files that live in parent directories) we use the `cmd/govim` prefix. 

Not ideal when mixing repos like this, but that's the current logic FWIW.

myitcv:
I'm going to merge this once #653 has landed.

PR #627, opened 2019-12-17
cmd/govim: fix scripts-with-gopls-from-path tests

myitcv:


myitcv:
Nothing controversial here; we're effectively turning some old tests back on :)

PR #636, opened 2019-12-20
testdriver: limit listening to local interface

leitzler:
Running the tests in macOS will bring up a dialog from the firewall
prompting to allow an unsigned process to listen to network connections.

Using local loopback interface instead for the testdriver will please the
firewall and no dialog shows.


myitcv:
Test failure is a new issue in `gopls`; raising as an issue shortly

PR #640, opened 2019-12-21
deps: upgrade x/tools and gopls to f13409bb

myitcv:


myitcv:
Failures are known test flakes. No changes in here from a `govim` perspective, so merging.

PR #643, opened 2019-12-21
cmd/govim: allow user defined signs

### LINKED ISSUE #633: cmd/govim: make quickfix gutter sign customizable
As far as I can tell, the quickfix sign is locked to `>>` in govim: https://github.com/govim/govim/blob/master/cmd/govim/signs.go#L44. I would love to be able to customize this.

(If it is customizable, I apologize! I looked hard for docs that suggested it was and even took a shot at tracing the code a bit. Also I elided all of the bug reporting sections of this issue because I'm not reporting a bug)

leitzler:
Currently govim defines the different sign types used at startup,
with hard coded sign (">>").

This adds a check before each sign definition that skips definition
if a sign with that name is already defined. It allows the user to
override sign types used by govim in their vimrc and that way use
custom signs.

e.g. to change sign to "!!" instead for errors:

call sign_define("GOVIMSignErr",{"text":"!!","texthl":"GOVIMSignErr"})

Fixes #633

### DIFF
--- a/cmd/govim/signs.go
+++ b/cmd/govim/signs.go
@@ -27,13 +27,40 @@ type defineDict struct {
 
 // signDefine defines the sign types (sign names) and must be called once before placing any signs
 func (v *vimstate) signDefine() error {
-	for _, hi := range []config.Highlight{config.HighlightSignErr, config.HighlightSignWarn, config.HighlightSignInfo, config.HighlightSignHint} {
+	signnames := []config.Highlight{
+		config.HighlightSignErr,
+		config.HighlightSignWarn,
+		config.HighlightSignInfo,
+		config.HighlightSignHint,
+	}
+	var useDefault []config.Highlight
+
+	// The user might have defined a sign name already in their vimrc, and govim should respect
+	// that and not override it.
+	v.BatchStart()
+	for _, hi := range signnames {
+		v.BatchChannelCall("sign_getdefined", string(hi))
+	}
+	for i, res := range v.BatchEnd() {
+		var d []defineDict
+		v.Parse(res, &d)
+		if len(d) == 0 {
+			useDefault = append(useDefault, signnames[i])
+		}
+	}
+
+	// Define default sign names
+	v.BatchStart()
+	for _, hi := range useDefault {
 		arg := defineDict{
 			Text:          ">>",
 			TextHighlight: string(hi),
 		}
 
-		if v.ParseInt(v.ChannelCall("sign_define", hi, arg)) != 0 {
+		v.BatchChannelCall("sign_define", hi, arg)
+	}
+	for _, res := range v.BatchEnd() {
+		if v.ParseInt(res) != 0 {
 			return fmt.Errorf("sign_define failed")
 		}
 	}


llimllib:
LGTM. To test, I:

1. switched to `~/.vim/plugged/govim`
2. checked out your branch (`hub checkout https://github.com/govim/govim/pull/643`)
3. Added to my `.vimrc`:
```
call sign_define("GOVIMSignErr",{"text":"🔥","texthl":"GOVIMSignErr"})
call sign_define("GOVIMSignWarn",{"text":"⚠️","texthl":"GOVIMSignWarn"})
call sign_define("GOVIMSignInfo",{"text":"⁉️","texthl":"GOVIMSignInfo"})
call sign_define("GOVIMSignHint",{"text":"⁉️","texthl":"GOVIMSignHint"})
```

After doing that, I see the updated signs when I open a file. 👍

llimllib:
(I don't know if I ought to approve the PR as I'm not a contributor to this repo? I'm going to assume not, but it has my approval)

leitzler:
Thanks for testing it out! I'll let @myitcv add his thoughts as well :)

cmd/govim/signs.go
```diff
@@ -27,13 +27,40 @@ type defineDict struct {
 
 // signDefine defines the sign types (sign names) and must be called once before placing any signs
 func (v *vimstate) signDefine() error {
-	for _, hi := range []config.Highlight{config.HighlightSignErr, config.HighlightSignWarn, config.HighlightSignInfo, config.HighlightSignHint} {
+	signnames := []config.Highlight{
```
myitcv: Not sure this will work if the user makes a call to:

```
call sign_define("GOVIMSignErr",{"text":"!!","texthl":"GOVIMSignErr"})
```

after startup? Arguably that will happen less often, but we should cover that case too.

i.e. we need to hook something into `*vimstate.setConfig` to have this working 100%

myitcv:
Thanks for working on this @leitzler - and thanks for prompting the suggestion @llimllib

cmd/govim/signs.go
leitzler: It does actually work already, re-defining an existing sign will overwrite, so that the new one is used instead.

myitcv: Yes but you need something like an edit after changing the sign to see it take effect. Hence I think we need a hook in `*vimstate.setConfig` to trigger replacing of the signs, no?

leitzler: Indeed, moving the cursor is enough to see it take effect. Is it possible for `govim` to detect such call? Since it isn't a part of `govim`'s configuration at all (it's all handled entirely in the context of vim).

I could raise an issue towards vim, like:
>  Existing signs isn't updated after a `sign_define()` call

myitcv: > Indeed, moving the cursor is enough to see it take effect.

Is it? In my tests locally a change was required in the file itself. Perhaps my test was bad?

leitzler: Hm, yeah. At least in Linux using Vim 8.2.0032 via mosh and macOS using Vim 8.2 (0000).

I open up a file in a temp directory (with go.mod):
```
package main

func main() {
	foo
}
```
Then call:
```
:call sign_define("GOVIMSignErr", {"text":"aa","texthl":"GOVIMSignErr"})
```
And move the cursor one step with the arrow keys.

myitcv: That worked! No idea what I did before. 

Still, I think we should fix it so that the user isn't left wondering why nothing changed on screen, would you agree?

leitzler: Yeah, agree fully. I’m not sure however if it’s possible to hook on to a ordinary call in vim. I haven’t seen that it generates any event, as it is handled outside of govim entirely.

We could create a govim config that “should” be used instead, but I don’t think that this scenario is common enough so that it is worth it. A user would still be able to call sign_define and run into the same issue.

Any other ideas how we could hook on to sign_define? Is it possible to override & wrap the ordinary one with a refresh? I’ll give it a try later tonight unless you already know.

myitcv: Good points all round :)

This actually smells more like a Vim issue than one for us.

Let's leave it as you have it for now on the `govim` side and then raise an issue with `vim_dev`?

leitzler: :+1: I rebased this PR against latest master & wrote https://github.com/vim/vim/issues/5455

PR #646, opened 2019-12-28
cmd/govim: tighten up quickfix config tests

myitcv:
In light of recent changes in gopls to not sent diagnostics for files
when those diagnostics have not changed, we need to slightly tweak our
assertions in various testscript scripts.

This is first such change. Others will/may follow

### DIFF
--- a/cmd/govim/testdata/scenario_default/quickfix_config.txt
+++ b/cmd/govim/testdata/scenario_default/quickfix_config.txt
@@ -3,20 +3,21 @@
 # Default behaviour is quickfix autodiagnostics & sign placment enabled
 vim ex 'e main.go'
 errlogmatch 'PublishDiagnostics callback: &protocol.PublishDiagnosticsParams{\n\S+:\s+URI:\s+"file://'$WORK/main.go
+errlogmatch -peek 'sendJSONMsg: \[0,\[\d+,"call","s:batchCall",\[\["call","s:mustNothing","setqflist"'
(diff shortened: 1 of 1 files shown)


cmd/govim/testdata/scenario_default/quickfix_config.txt
```diff
@@ -3,20 +3,21 @@
 # Default behaviour is quickfix autodiagnostics & sign placment enabled
 vim ex 'e main.go'
 errlogmatch 'PublishDiagnostics callback: &protocol.PublishDiagnosticsParams{\n\S+:\s+URI:\s+"file://'$WORK/main.go
+errlogmatch -peek 'sendJSONMsg: \[0,\[\d+,"call","s:batchCall",\[\["call","s:mustNothing","setqflist"'
```
leitzler: Do we need `-peek` here? IIRC errlogmatch doesn't consume any lines after a matched line anymore.
If we do need it, shouldn't it be added to the PublishDiagnostics match above as well?

myitcv: I don't think we do, no. Reason being, we know the diagnostics notification will come from `gopls` first. Then, without attaching too much to the current implementation (which is a bit of a moot point because that's exactly what we are doing) we then say the quickfix update and signs update can happen in either order. In reality, they will always happen in the same order, but the `-peek` is really to loosen that a bit. But all of these assertions are so tightly coupled to the implementation it's really not worth it in some respects.

leitzler: I see, good point 👍

PR #650, opened 2020-01-03
tests: create gists from artefacts (logs) of failed CI tests

myitcv:


cmd/govim/main.go
```diff
@@ -456,3 +467,12 @@ func (g *govimplugin) defineHighlights() {
 	}
 	g.vimstate.BatchEnd()
 }
+
+func getEnvVal(env []string, varname string) (string, bool) {
```
leitzler: Nits, is the bool really needed? It isn't used at all in this PR

myitcv: Thanks, good catch. I forgot to refactor this.

PR #651, opened 2020-01-04
ci: fix simple bug in dockerRun script and Travis config

myitcv:


myitcv:
Test failures are known flakes. Will be fixed in later PRs in this stack

Issue #658, opened 2020-01-08
cmd/govim: recent x/tools, gopls upgrade emits extra junk in omnicompletion


cespare:
After 11eb9c75ea607c4caf63a93e2d8016b2c0099c8f (#652, "deps: upgrade x/tools and gopls to 774c71fc") omnicompletion includes some extra markup text:

```
	mac := hmac.New(sha256.New, []byte(key))
	mac.Su<hit C-x C-o or whatever to omnicomplete here>
```

yields

```
	mac := hmac.New(sha256.New, []byte(key))
	mac.Sum(${1:})
```

Before that change, it would just complete the name of the function:

```
	mac := hmac.New(sha256.New, []byte(key))
	mac.Sum
```

cespare:
Ugh, hit enter too early. Editing in details now...

myitcv:
Thanks for the report. 

I spotted this last night but was hoping I'd be able to squeeze in a fix before anyone noticed.

The back story is:

* we upgrade `x/tools` and `gopls` in 11eb9c75ea607c4caf63a93e2d8016b2c0099c8f
* I spotted the bug immediately and Rebecca quickly raised and merged https://go-review.googlesource.com/c/tools/+/213640 as a fix
* the problem is that in the intervening time https://go-review.googlesource.com/c/tools/+/212102 had been merged which introduced another bug
* https://go-review.googlesource.com/c/tools/+/213641 was quickly raised to fix that but it transpires the fix will take long

Hence I've now:

* raised https://github.com/govim/govim/pull/659 to test coverage of this sort of completion
* raised https://github.com/govim/govim/pull/660 to upgrade `x/tools` and `gopls` to a temp fork where this bug is fixed

cespare:
Thanks for the info! And no worries, I expect to encounter stuff like this when I keep using master.

myitcv:
We shouldn't really have this sort of stuff in master though to my mind. Because most Vim plugin managers/approaches effectively default to using master (as opposed to released versions) we need to be a bit more careful.

PR #659, opened 2020-01-08
cmd/govim: create a test demonstrating failing completion

myitcv:
As previously discovered (and also raised in #658), 11eb9c7 caused
completion to break in govim because placeholder values are now included
in the completion result. This has now been fixed in:

  https://go-review.googlesource.com/c/tools/+/213640

However, because of a bug introduced in:

  https://go-review.googlesource.com/c/tools/+/212102

we can't simply upgrade to the latest x/tools and gopls.

In any case we want to add a test to demonstrate the problem and prevent
and future regressions. A subsequent PR will upgrade the version of
x/tools and gopls to a commit (in a temporary fork) that has the fix for
the completion bug but does not have the bug for the parallel initial
load.

Updates #658


myitcv:
This will send the entire build red. But #660 will fix it. So I will merge this once the build for #660 is green.

PR #660, opened 2020-01-08
deps: use temp fork of x/tools and gopls to 53017a39 + 11e9d9cc

myitcv:


myitcv:
Given there are no changes required in `govim` here, merging.

myitcv:
The build failures are known flakes that will be fixed by #653

PR #663, opened 2020-01-08
[DO NOT MERGE] cmd/govim: ignore non-new diagnostics

myitcv:
In case gopls sends non-new diagnostics for a file we should ignore then
(and return an error to gopls). This is effectively a fail-safe on the
govim side that shouldn't really be necessary.


leitzler:
Related/blocked by https://github.com/golang/go/issues/36476

myitcv:
Per https://github.com/golang/go/issues/36601#issuecomment-575627675 we will very likely be dropping this PR. So will mark as `[DO NOT MERGE]`

myitcv:
Closing per https://github.com/govim/govim/pull/663#issuecomment-575652909

Issue #655, opened 2020-01-06
cmd/govim: investigate prop_add flake


myitcv:
From a recent CI run. The test itself doesn't appear to be significant, except for the fact that line removal is probably involved:

```
--- FAIL: TestScripts (43.60s)
    main_test.go:63: using gopls at "/tmp/gobin-gopls-installdir074416177/gopls"
    --- FAIL: TestScripts/scenario_default (0.00s)
        --- FAIL: TestScripts/scenario_default/goimports (1.43s)
            testscript.go:337: 
                # Test that govim#config#Set("FormatOnSave", "goimports") and GOVIMGoImports work
                # Notice below: file.go and file.go.orig are identical. This is so that we
                # don't need to copy file.go.orig to file.go before opening file.go Because
                # doing so triggers govim to consider the file change from a file watcher
                # perspective which throws the observed diagnostics etc.
                # :GOVIMGoImports whole file (0.594s)
                # Format on save (0.313s)
                # Disabled pending resolution to https://github.com/golang/go/issues/34103
                # errlogmatch -count=0 'LogMessage callback: &protocol\.LogMessageParams\{Type:(1|2), Message:".*'
                # Format on save (bad syntax) (0.045s)
                > cp file.go.bad file.go
                > vim ex 'e! file.go'
                [stderr]
                unexpected command error: failed to ex(e! file.go) in Vim: Caught 'got error whilst handling autocommand:1: driver error: ChannelCall("s:batchCall") failed: failed to call s:batchCall([[[call s:mustNothing prop_remove {0 1 1}] [call s:mustNothing prop_add 4 5 {GOVIMErr 4 21 1}]]]) in Vim: Caught ''Vim(let):E966: Invalid line number: 4'' in function <SNR>22_callbackAutoCommand[18]..<SNR>22_ch_evalexpr[2]..<SNR>22_define[51]..<SNR>22_batchCall, line 21' in function <SNR>22_callbackAutoCommand[18]..<SNR>22_ch_evalexpr, line 4
                [exit status 255]
                FAIL: testdata/scenario_default/goimports.txt:29: unexpected command failure
                
FAIL
FAIL	github.com/govim/govim/cmd/govim	149.698s
```

i.e. we try to add props to a line that no longer exists?

myitcv:
Logs available from a recent build fail via:

```
[...]

PR #665, opened 2020-01-10
cmd/govim: Do not increase version if buffer content didn't change

leitzler:
We currently blindly increase buffer version when vim triggers
BufReadPost or BufNewFile (and a didOpen or didChange request to
gopls). Unloading and loading a buffer that is open also triggers
a BufReadPost, and if the content didn't change we shouldn't
increase version.

This change also skip redefining highlights if the version is
indeed increased since the last known diagnostics could have
changed.

Updates #655

### DIFF
--- a/cmd/govim/buffer_events.go
+++ b/cmd/govim/buffer_events.go
@@ -28,8 +28,18 @@ func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
 
 	if cb, ok := v.buffers[nb.Num]; ok {
 		// reload of buffer, e.v. e!
-		cb.SetContents(nb.Contents())
 		cb.Loaded = nb.Loaded
+
+		// If the contents are the same we probably just re-loaded a currently unloaded buffer.
+		// We shouldn't increase version in that case, but we have to redefine highlights
+		// since text properties are removed when a buffer is unloaded.
+		if bytes.Equal(nb.Contents(), cb.Contents()) {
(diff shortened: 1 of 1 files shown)


cmd/govim/buffer_events.go
```diff
@@ -28,8 +28,19 @@ func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
 
 	if cb, ok := v.buffers[nb.Num]; ok {
 		// reload of buffer, e.v. e!
-		cb.SetContents(nb.Contents())
 		cb.Loaded = nb.Loaded
+
+		// If the contents is the same we probably just re-loaded an already open buffer.
```
myitcv: "If the contents _are_ the same..."

I don't think it's strictly correct to say "open buffer" either. Would it perhaps be more accurate to say "... probably just re-loaded a currently unloaded buffer"?

myitcv:
FWIW the tests that we do add around this probably belong alongside the diagnostics expectation tests:

https://github.com/govim/govim/tree/4208a9805df15a50ba6e714ca806d425351d50bd/cmd/govim/testdata/scenario_diagnostic_assertions

leitzler:
Agree, raised #667 as a reminder.

cmd/govim/buffer_events.go
leitzler: Good point, updated!

PR #671, opened 2020-01-10
cmd/govim: change log file template to default to .log files

myitcv:
A minor annoyance, but when reporting issues via GitHub, uploading log
files only works in the browser if the file has a .log extension. Our
log file templates do not currently have such an extension which means
manual renaming is required before the file can be uploaded.

Fix that.


myitcv:
Thank you - that's a good catch.

myitcv:
I ended up having to test this locally to ensure all is fixed. Now done so will merge.

PR #674, opened 2020-01-13
cmd/govim: apply AdditionalTextEdits on CompleteDone

myitcv:
Some completion candidates, most notably unimported completion
candidates, carry with them additional text edits. In the case of
unimported candidates, these represent the changes required to add the
missing import if the candidate is chosen.

This commit applies those text edits when we get the event CompleteDone.

### DIFF
--- a/cmd/govim/complete.go
+++ b/cmd/govim/complete.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/govim/govim"
 	"github.com/govim/govim/cmd/govim/internal/golang_org_x_tools/lsp/protocol"
 	"github.com/govim/govim/cmd/govim/internal/types"
 )
@@ -47,13 +48,14 @@ func (v *vimstate) complete(args ...json.RawMessage) (interface{}, error) {
 		v.lastCompleteResults = res
 		return start, nil
 	} else {
-		var matches []completionResult
+		var matches []govim.CompleteItem
 		for _, i := range v.lastCompleteResults.Items {
-			matches = append(matches, completionResult{
-				Abbr: i.Label,
-				Menu: i.Detail,
-				Word: i.TextEdit.NewText,
-				Info: i.Documentation,
+			matches = append(matches, govim.CompleteItem{
+				Abbr:     i.Label,
+				Menu:     i.Detail,
+				Word:     i.TextEdit.NewText,
+				Info:     i.Documentation,
+				UserData: "govim",
 			})
 		}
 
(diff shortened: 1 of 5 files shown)


cmd/govim/complete.go
```diff
@@ -61,6 +61,35 @@ func (v *vimstate) complete(args ...json.RawMessage) (interface{}, error) {
...
+		return nil
+	}
+	var match *protocol.CompletionItem
+	for _, c := range v.lastCompleteResults.Items {
+		if c.Label == chosen.Abbr {
+			match = &c
+		}
+	}
+	if match == nil {
+		return fmt.Errorf("failed to find match for completed item %#v", chosen)
```
leitzler: This might actually happen in a OK state, if someone performs a content-based completion (after, at some point, having performed a omnicomplete to prevent the panic above).

cmd/govim/complete.go
```diff
@@ -61,6 +61,35 @@ func (v *vimstate) complete(args ...json.RawMessage) (interface{}, error) {
...
+	var chosen struct {
+		Word string `json:"word"`
+		Abbr string `json:"abbr"`
+	}
+	v.Parse(args[1], &chosen)
+	if chosen.Word == "" {
+		return nil
+	}
+	var match *protocol.CompletionItem
+	for _, c := range v.lastCompleteResults.Items {
```
leitzler: This will panic with a nil-deref if the user performs a content-based completion (C-x C-n) without performing a omnicomplete before.

myitcv:
Thanks for the catch on the `<C-x><C-n>`-type completions. We might be able to address the points you raise slightly differently based on any responses we get to:

https://groups.google.com/forum/#!topic/vim_dev/z7acHIqGMBE

Otherwise we might need to be a bit more defensive in our code, as you suggest.

myitcv:
Marking this as onhold until we hear back on the Vim question.

cmd/govim/complete.go
```diff
@@ -61,6 +61,35 @@ func (v *vimstate) complete(args ...json.RawMessage) (interface{}, error) {
 	}
 }
 
+func (v *vimstate) completeDone(args ...json.RawMessage) error {
+	currBufNr := v.ParseInt(args[0])
+	b, ok := v.buffers[currBufNr]
+	if !ok {
+		return fmt.Errorf("failed to resolve buffer %v", currBufNr)
+	}
+	var chosen struct {
```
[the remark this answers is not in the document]
myitcv: Thanks, good catch. I've also moved these types to `github.com/govim/govim` because they're generically useful.

cmd/govim/complete.go
myitcv: Thanks, per our Slack chat I've added logic per Bram's response here https://groups.google.com/d/msg/vim_dev/z7acHIqGMBE/GoGpGnGzAwAJ

cmd/govim/complete.go
myitcv: Resolved per previous point.

myitcv:
All changes made: PTAL

PR #677, opened 2020-01-15
scripts: add captureLogs.sh to easily run tests and capture all logs

myitcv:
When the CI build fails and generates artefacts we capture logs.
Sometimes it's useful to be able to do that locally.

_scripts/captureLogs.sh makes that trivial by wrapping the relevant
commands.

### DIFF
--- a/_scripts/captureLogs.sh
+++ b/_scripts/captureLogs.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+source "${BASH_SOURCE%/*}/common.bash"
+source "${BASH_SOURCE%/*}/tidyUp.bash"
+
+if [[ $# -eq 0 ]] ; then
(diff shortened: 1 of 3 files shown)


_scripts/captureLogs.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# captureLogs.sh is used to simplify the capture of all logs from a run
+# of tests. It takes the following arguments:
+#
+# captureLogs.sh dir command args...
+
+source "${BASH_SOURCE%/*}/common.bash"
+source "${BASH_SOURCE%/*}/tidyUp.bash"
+
+dir="$1"
```
leitzler: Nits, a usage dump if started w/o argument would be nice.

myitcv: Great suggestion; done.

PR #678, opened 2020-01-15
testdriver: add vimexprwait to simplify wait assertions

myitcv:


myitcv:
@findleyr would appreciate any comments/thoughts you have on this.

Notice that at the time of writing, this PR is stacked atop #677 hence you need to look at the last commit under https://github.com/govim/govim/pull/678/commits

testdriver/testdriver.go
```diff
@@ -149,27 +150,13 @@ func NewTestDriver(c *Config) (*TestDriver, error) {
...
-	// Instead below we copen then cclose in vimrc to reserve and establish a
-	// definite buffer number for the quickfix window. The quickfix buffer will
-	// always be buffer number 2 (vim creates a buffer when opening, copen then
-	// creates a new window and buffer, the original buffer created when opening
-	// vim gets used for the first file we open, say)
 	vimrcBytes, err := ioutil.ReadFile(srcVimrc)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read from src vimrc %v: %v", srcVimrc, err)
 	}
 	vimrc := bytes.NewBuffer(vimrcBytes)
```
findleyr: Optional: skip the buffer since you are no longer augmenting the vimrc.

testdriver/testdriver.go
```diff
@@ -538,46 +525,148 @@ func (d *TestDriver) listenDriver() error {
...
+		// this is a protocol-level error
+		ef("got error response: %v", resp[0])
 	}
-	log("logging enabled")
-
-	defer func() {
-		r := recover()
-		if r == nil {
-			return
+	// resp[1] will be a []intferface{} where the first
```
findleyr: typo: interface

testdriver/testdriver.go
```diff
@@ -538,46 +525,148 @@ func (d *TestDriver) listenDriver() error {
...
+// JSON. resp is a tuple such that:
+//
+// resp[0] is a string representing any protocol error
+// resp[1] is a []interface such that:
+//   resp[1][0] is a string representing any Vim error
+//   resp[1][1] (if supplied) is the value returned by Vim
+func vim(args []string) (resp [2]interface{}, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = r.(error)
```
findleyr: What if r is not an error, as would be the case from ef?

testdriver/testdriver.go
```diff
@@ -538,46 +525,148 @@ func (d *TestDriver) listenDriver() error {
...
+	var got *bytes.Buffer
+	for a := retry.Start(strategy, nil); a.Next(); {
+		resp, err := vim(vimCmd)
+		if err != nil {
+			ef("%v", err)
+		}
+		if resp[0] != "" {
+			ef("protocol error: %v", resp[0])
+		}
+		vimResp := resp[1].([]interface{})
```
findleyr: Not sure if this is possible, but could you push down this type coercion into the vim func, and return a better type?

_scripts/tidyUp.bash
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+source "${BASH_SOURCE%/*}/common.bash"
+
+tidyUp() {
+	# Remove all the big directories first
+	sudo find "$1" -type d \( -name .vim -o -name gopath \) -prune -exec rm -rf '{}' \;
```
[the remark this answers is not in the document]
myitcv: The 0777 permissions are before umask; most distros have a default umask of 0022 hence we end up with 0755 for the directory permissions.

The reason the sudo is required here is that GOPATH directory structures will be contained therein, and the GOPATH/pkg/mod directory structure in particular has read-only files and directories. Hence we need sudo to remove those directories.

testdriver/testdriver.go
myitcv: I was actually trying to be forward thinking in leaving in a convenient placeholder for adding augmentations to the vimrc. But it's not really work it so I'll switch to using `copyFile` as elsewhere because that's all we need here.

testdriver/testdriver.go
myitcv: Thanks

testdriver/testdriver.go
myitcv: Thanks good catch. I missed changed `ef` to panic with `fmt.Errorf`

_scripts/tidyUp.bash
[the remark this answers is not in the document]
findleyr: Ack.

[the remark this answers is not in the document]
myitcv: I actually switched to dropping the use of sudo and instead using chmod -R to add user write perms. Seemed safer than using the nuclear sudo option per your comments

testdriver/testdriver.go
```diff
@@ -538,46 +515,145 @@ func (d *TestDriver) listenDriver() error {
...
+		}
+	}
+	// We failed to find a match. Show diff and return an error
+	fmt.Fprint(os.Stderr, textutil.Diff(got.String(), string(want)))
+
+	return 1
+}
+
+type vimResponse struct {
+	protocolError string
```
findleyr: I think you could probably remove protocolError, instead returning an error in the second return value in these cases. You only ever do
if err != nil {
  ef(...)
}
if resp.protocolError != nil {
  ef(...)
}
so these can be rolled together.

testdriver/testdriver.go
```diff
@@ -538,46 +515,145 @@ func (d *TestDriver) listenDriver() error {
...
+	vimError      string
```
findleyr: Optional, but I'd make this a proper error, for readability and so it can be an arbitrary value.

testdriver/testdriver.go
myitcv: Thanks, that's a nice simplification. Done.

testdriver/testdriver.go
myitcv: Thanks - yep with you on this too.

myitcv:
Known flake in the tests. Also discussed this PR with @leitzler offline and we agreed it's a 👍

PR #683, opened 2020-01-16
cmd/govim: add diagnostics to hover popup

### LINKED ISSUE #468: cmd/govim: show error details when hovering over diagnostic error
Per suggestion from @julieqiu; we should default to showing a hover popup when the mouse moves over a diagnostic error. An error-related popup takes higher priority than the existing hover results.

As part of this we should also ensure that the error range is highlighted (we get the exact range details back in the diagnostic from `gopls` so this should be easy). In GVim this highlighting should likely use the squiggly underline.

cc @leitzler for thoughts, given the link with the signs work.

leitzler:
When using the hover feature, either via keyboard or mouse, any
diagnostics with a range that covers the current position will
be added at the top of popup.

The diagnostics are formatted using text properties, and as a
consequence of this there are a few new highlight groups defined:

* GOVIMHoverErr
* GOVIMHoverWarn
* GOVIMHoverInfo
* GOVIMHoverHint

The diagnostic line also includes the diagnostic source (as
provided by gopls, e.g. "compiler", "printf", "unreachable").
To be able to format the source there is another highlight group
used for the source part:

* GOVIMHoverDiagSrc

All text properties are combined with existing syntax highlight
to enable a wide range of custom styles.

Closes #468

### DIFF
--- a/autoload/govim/config.vim
+++ b/autoload/govim/config.vim
@@ -68,6 +68,10 @@ function! s:validHighlightDiagnostics(v)
     return s:validBool(a:v)
 endfunction
 
+function! s:validHoverDiagnostics(v)
+    return s:validBool(a:v)
+endfunction
+
 function! s:validCompletionDeepCompletions(v)
   return s:validBool(a:v)
 endfunction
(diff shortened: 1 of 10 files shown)


leitzler:
Inspired by VSCode that shows diagnostics (incl. source) in the existing hover function (e.g. documentation) this acts as a first cut.

E.g.: [asciicast](https://asciinema.org/a/8zvvqBGDNi5z9m6EEUh8C7CSS)

Example from VSCode:
image

myitcv:
This is *very* cool.

myitcv:
Reminder to selves from Slack conversation; we should provide a config option to disable this (or bundle that in an existing config option)

leitzler:
Closed in favour of #693

PR #689, opened 2020-01-19
deps: use temporary fork of x/tools and gopls

myitcv:
In d80f0e21 we upgraded to the latest x/tools and gopls. But as is
documented in golang.org/issue/36601 there is a bug whereby reverting a
file to an earlier state results in no diagnostics where they are in
fact expected.

Fix this by using a temporary fork of x/tools and gopls where CL 214586
is partially reverted.

Also add tests to ensure the various permutations of transitions to/from
error state and back work as expected.


cmd/govim/testdata/scenario_bugs/bug_gopls_36601_no-error_no-error_no-error.txt
```diff
@@ -0,0 +1,51 @@
...
+module mod.com
+
+go 1.12
+-- main.go --
+package main
+
+func main() {
+	//
+}
+-- errors.golden --
```
leitzler: `errors.golden` isn't used in this test

myitcv: Thanks, good spot. Wonder whether we could write a simple script to warn about potential instances of this kind?

PR #691, opened 2020-01-19
cmd/govim: make testscript scripts truly concurrent

myitcv:
Our cmd/govim testscript tests are split across a number of directories.
This grouping allows us to have separate configuration per directory.

Currently we do not call t.Parallel in each directory's subtest. This
means that we run directory scipts in serial, one directory after
another. Therefore, if there is a particularly slow test in one
directory, we can end up with flat spots where we do not fully utilise
the -test.parallel concurrency limit.

Making directory subtests parallel removes these flat spots. But we do
so at the cost of leaving beind gopls install artefacts. When we start
testing against a Go 1.14 release that includes CL 214822 we can enable
that cleanup, but Go version pre 1.14 will unfortunately continue to
leak these artefacts.

### DIFF
--- a/_scripts/runDockerRun.sh
+++ b/_scripts/runDockerRun.sh
@@ -8,8 +8,9 @@ cd "${BASH_SOURCE%/*}/../"
 
 # The ARTEFACTS variable set by .travis.yml cannot expand
 # variables so we do that here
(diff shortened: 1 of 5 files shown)


myitcv:
@leitzler ended up reviewing this commit as part of #326. So merging.

PR #693, opened 2020-01-19
cmd/govim: add diagnostics to hover popup

### LINKED ISSUE #468: cmd/govim: show error details when hovering over diagnostic error

leitzler:
When using the hover feature, either via keyboard or mouse, any
diagnostics with a range that covers the current position will
be added at the top of popup.

The diagnostics are formatted using text properties, and as a
consequence of this there are a few new highlight groups defined:

* GOVIMHoverErr
* GOVIMHoverWarn
* GOVIMHoverInfo
* GOVIMHoverHint

The diagnostic line also includes the diagnostic source (as
provided by gopls, e.g. "compiler", "printf", "unreachable").
To be able to format the source there is another highlight group
used for the source part:

* GOVIMHoverDiagSrc

All text properties are combined with existing syntax highlight
to enable a wide range of custom styles.

The feature is enabled by default, and can be disabled by setting
the config HoverDiagnostics to 0 (false).

Closes #468

### DIFF
--- a/autoload/govim/config.vim
+++ b/autoload/govim/config.vim
@@ -68,6 +68,10 @@ function! s:validHighlightDiagnostics(v)
     return s:validBool(a:v)
 endfunction
 
+function! s:validHoverDiagnostics(v)
+    return s:validBool(a:v)
+endfunction
+
 function! s:validCompletionDeepCompletions(v)
   return s:validBool(a:v)
 endfunction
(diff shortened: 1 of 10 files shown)


leitzler:
This PR supersedes #683 to ensure Travis builds works as expected.

leitzler:
As per my comment on Slack - the lower part of the popup might be chopped off if there is a lot of content (e.g. the http.Server struct). If we add the diagnostics to the bottom part they might not be visible at all.

I'll keep it like this as a starting point, since it's better to drop a few lines docs than the actual diagnostics.

PR #699, opened 2020-01-20
cmd/govim: add testscript tests for new files not on disk

myitcv:


myitcv:
This will fail because of https://github.com/golang/go/issues/36661

Issue #685, opened 2020-01-17
cmd/govim: place signs and add text properties in try/catch


myitcv:
Adding a placeholder issue to keep track of conversation in:

https://groups.google.com/forum/#!topic/vim_dev/tV_iEQL0rOE

> In govim we get diagnostics about code that is being edited in Vim via
async notifications from gopls, the Go LSP server implementation. We
then place signs and add text properties as a function of those
diagnostics to highlight issues to the user.
>
> Diagnostics have a versioned reference to the file to which they correspond.
>
> As an efficiency measure, gopls only resends diagnostics for a file if
the resulting diagnostics change. e.g. making a change to a file at
version v (resulting in v') will only trigger new diagnostics to be
sent for that file (at version v') if they (the diagnostics) are
different to the diagnostics for version v. The justification for this
approach is performance and based on evidence thus far it's a sound
decision.
>
> However this has the implication that govim cannot use its state of
the file's version to only apply diagnostics (place signs, add text
properties) for that file when it matches the diagnostic's versioned
reference. Because the diagnostics for v' might well still be current
for file version v''.
>
> Hence we need to place signs/add text properties in such a way that we
know they might fail (because it might be the case that we have simply
not yet received the latest "correct" diagnostics for that file -
diagnostics are, after all, async).
>
> One approach we are therefore considering is to place signs/add text
properties within a try/catch block, and simply swallow any errors
safe in the knowledge that "things will right themselves soon").
>
> Does anyone foresee any problems with this approach?
>
> Is there anything else we should be considering here?

cc @leitzler

leitzler:
Close by #700 & #741

PR #700, opened 2020-01-20
cmd/govim: leaving userbusy always updates text property highlighs

leitzler:
The diagnostic highlights were always updated when leaving userbusy.
If the buffer changed since the last diagnostic received from gopls,
we might end up with a buffer that doesn't contain the lines/cols
where we try to place text properties. That did thrown an error.

This fix adds the same logic that exists for other diagnostic
features, so that leaving userbusy only triggers text property
placment if there are new diagnostics.

### DIFF
--- a/cmd/govim/buffer_events.go
+++ b/cmd/govim/buffer_events.go
@@ -34,7 +34,7 @@ func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
 		// We shouldn't increase version in that case, but we have to redefine highlights
 		// since text properties are removed when a buffer is unloaded.
 		if bytes.Equal(nb.Contents(), cb.Contents()) {
-			if err := v.redefineHighlights(v.diagnostics()); err != nil {
+			if err := v.redefineHighlights(v.diagnostics(), true); err != nil {
 				v.Logf("failed to update highlights for buffer %d: %v", nb.Num, err)
 			}
 			return nil
@@ -61,7 +61,7 @@ func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
 		v.Logf("failed to update signs for buffer %d: %v", nb.Num, err)
 	}
 
-	if err := v.redefineHighlights(v.diagnostics()); err != nil {
+	if err := v.redefineHighlights(v.diagnostics(), true); err != nil {
 		v.Logf("failed to update highlights for buffer %d: %v", nb.Num, err)
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/govim/buffer_events.go
```diff
@@ -61,7 +61,7 @@ func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
 		v.Logf("failed to update signs for buffer %d: %v", nb.Num, err)
 	}
 
-	if err := v.redefineHighlights(v.diagnostics()); err != nil {
+	if err := v.redefineHighlights(v.diagnostics(), true); err != nil {
```
myitcv: I'm not sure this is guaranteed to be safe. Because it's possible a cheeky file change snuck in before we opened the file which would make the current diagnostics for that file "bad". Calling with `true` is correct, but per our Slack chat this is one situation where I think we could end up applying bad highlights.

cmd/govim/gopls_client.go
```diff
@@ -179,6 +179,7 @@ func (g *govimplugin) PublishDiagnostics(ctxt context.Context, params *protocol.
 	g.diagnosticsChanged = true
 	g.diagnosticsChangedQuickfix = true
 	g.diagnosticsChangedSigns = true
+	g.diagnosticsChangedHighlights = true
```
myitcv: At some point in the future I think it will make sense to shift this to by `uri` level granularity for whether diagnostics have changed, particularly when it comes to updating signs and highlights.

myitcv: Created https://github.com/govim/govim/issues/701 as a reminder

cmd/govim/vimstate.go
```diff
@@ -113,9 +113,9 @@ func (v *vimstate) setConfig(args ...json.RawMessage) (interface{}, error) {
 	if !vimconfig.EqualBool(v.config.HighlightDiagnostics, preConfig.HighlightDiagnostics) {
 		if v.config.HighlightDiagnostics == nil || !*v.config.HighlightDiagnostics {
 			// HighlightDiagnostics is now not on - remove existing text properties
-			v.redefineHighlights([]types.Diagnostic{})
+			v.redefineHighlights([]types.Diagnostic{}, true)
```
myitcv: Again this is another situation where I think we could end up trying to apply bad highlights. 

* User makes change
* User quickly changes config
* Diagnostics are the old diagnostics (new ones haven't arrived yet)
* Bad diagnostics get applied as highlights

leitzler: This particular case is safe since it's only removing diagnostics (passing in an empty slice).

cmd/govim/gopls_client.go
leitzler: Fully agree, thanks for creating the issue

cmd/govim/vimstate.go
myitcv: Ah yes, commented on wrong one.

cmd/govim/buffer_events.go
leitzler: Yeah, we'll handle that in https://github.com/govim/govim/issues/685.

PR #705, opened 2020-01-21
ci: notify when the master build goes green

myitcv:
Currently we never notify for a green build on master. This isn't very
helpful because it means we don't get a notification when the build
passes having previously failed, i.e. red -> green.

Fix that to notify on change.


myitcv:
This is a fair simple change so will merge when we get a green light!

PR #706, opened 2020-01-21
deps: upgrade x/tools and gopls to d456b1cd

myitcv:
* internal/imports: pass dummy source for completion functions d456b1cd
* internal/lsp: use correct file identities when computing diagnostics 62545537
* internal/lsp/cache: create infra for caching go.mod diagnostics bd79bb77
* internal/lsp/cache: fix mod file change check 9375b12b
* internal/lsp/protocol: remove unused DocumentUri type 13c74800
* internal/lsp/source: fix typeIsValid() inf recursion dbc83e6d
* go/ssa: remove workaround for lack of vendoring in go/loader eb0d8dd8
* go/ssa/interp: fix goroutines and channels leaks e919f867


myitcv:
There are no `govim` changes here so I will merge when the tests go green. 

This also has the benefit of removing our use of the temporary fork introduced in https://github.com/govim/govim/commit/c7a6cc371a422fa87adcbd422320a9955187ca4d

PR #712, opened 2020-01-25
cmd/govim: make all error/warning message checks identical

myitcv:
Ensure that all of the commented-out error/warning message checks at the
end of testscripts are identical and correct.

Do so by writing a simple helper program, txtarutil, that does this for
us. As part of this make genconfig an internal command along with
txtarutil. Both commands run as part of go:generate directives.

There should be no semantic changes to any tests as part of this commit,
only comments.


cmd/govim/testdata/logmessage.footer
```diff
@@ -0,0 +1,2 @@
+
+errlogmatch -start -count=0 'LogMessage callback: &protocol\.LogMessageParams\{Type:(1|2), Message:".*'
```
leitzler: Nits - Now that it's a template we could add a comment about what 1 & 2 represents so that a reader understands directly the purpose.

myitcv: Good point. Will fix that.

PR #714, opened 2020-01-25
ci: tidy up go.{mod,sum}

myitcv:
Remove unnecessary indirect module requirements


myitcv:
This is an uninteresting change so merging.

PR #715, opened 2020-01-25
testdriver: add common test functions

myitcv:


myitcv:
Merging because a3b392d and fce7545 are identical.

PR #721, opened 2020-01-26
cmd/govim: split config.FormatOnSaveGoImports into component parts

myitcv:
WIP


myitcv:
> (We should probably add a test for the default behaviour / goimports-gofmt as well.)

Done.

PR #727, opened 2020-01-27
cmd/govim: add tests for gopls features where files are initially empty

myitcv:
We have a number of tests that check gopls behaviour where files do not
exist on disk, i.e. they are just buffers in memory. But we are missing
tests to verify that if .go files on disk are empty and then "filled in"
by changes in the editor, that things work as expected.

Add some tests of that sort.

### DIFF
--- a/cmd/govim/testdata/scenario_default/quickfix_empty_files.txt
+++ b/cmd/govim/testdata/scenario_default/quickfix_empty_files.txt
@@ -1,0 +1,131 @@
+# Test that we get diagnostics for when there are empty files on disk but
+# in the editor we fill in the content.
+#
+# The scenario involves creating a package p with a simple function DoIt. p is
+# imported by a main package. p also has a test file that exercises DoIt, and
+# an external test file that does the same.
+#
+# Initially all call sites for DoIt incorrectly pass an integer argument,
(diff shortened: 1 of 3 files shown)


cmd/govim/testdata/scenario_default/quickfix_empty_files.txt
```diff
@@ -0,0 +1,133 @@
...
+# The scenario involves creating a package p with a simple function DoIt. p is
+# imported by a main package. p also has a test file that exercises DoIt, and
+# an external test file that does the same.
+#
+# Initially all call sites for DoIt incorrectly pass an integer argument,
+# meaning we should have error diagnostics for all call sites. Then we correct
+# the definition of DoIt to take an integer argument at which point all
+# diagnostics should disappear.
+
+# skip 'Pending fix for golang.org/issues/36795'
```
leitzler: I guess this is commented out to provide details to Rebecca & co(?). If so, don't forget to uncomment before merge (or remove if it's fixed).

myitcv: Done

myitcv:
Mac builds seem to be taking an inordinate amount of time. Merging.

PR #728, opened 2020-01-27
cmd/govim: add go-to-def tests for std lib and module cache definitions

myitcv:
Extend our suite of go-to-def tests by adding two new tests that jump to
standard library and module cache definitions.


cmd/govim/testdata/scenario_default/go_to_def_module_cache.txt
```diff
@@ -0,0 +1,37 @@
+# Test that GOVIMGoToDef works jumping to a definition within
+# the module cache
+
+# More natural to split below and to the right
```
leitzler: Do splits matter for the test(s)?

PR #730, opened 2020-01-27
cmd/govim: Ensure text prop highlights aren't placed when disabled

### LINKED ISSUE #724: cmd/govim: Disabling HighlightDiagnostics sometimes doesn't work
I'm on the latest master as of right now (4b70552ac41586b1cd509e72828bd67236292133). This may be a recent regression since I don't recall this happening until the last few days.

In my .vimrc I have

    call govim#config#Set("HighlightDiagnostics", 0)

Weirdly, I occasionally get the diagnostic highlights (red underline squiggles) anyway.

Unfortunately, I don't know how to reproduce this. Sometimes I use vim for an hour before it starts happening. Once I got it to happen right after I launched vim. I don't even know if, once it starts happening, it keeps happening -- each time so far when it has happened I've closed vim right away. I'll update this issue when I figure out more.

leitzler:
The config HighlightDiagnostics can be used to disable in-code
highlights. PR #700 accidently broke that feature so that it was
possible to get highlights even when disabled.

This PR ensures that text prop highlights aren't placed when they
are disabled by the config.

Fixes #724

### DIFF
--- a/cmd/govim/highlight.go
+++ b/cmd/govim/highlight.go
@@ -51,7 +51,7 @@ func (v *vimstate) textpropDefine() error {
 }
 
 func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) error {
-	if !force && (v.config.HighlightDiagnostics == nil || !*v.config.HighlightDiagnostics) {
+	if v.config.HighlightDiagnostics == nil || !*v.config.HighlightDiagnostics {
 		return nil
 	}
 	v.diagnosticsChangedLock.Lock()
--- a/cmd/govim/vimstate.go
+++ b/cmd/govim/vimstate.go
@@ -113,7 +113,18 @@ func (v *vimstate) setConfig(args ...json.RawMessage) (interface{}, error) {
 	if !vimconfig.EqualBool(v.config.HighlightDiagnostics, preConfig.HighlightDiagnostics) {
 		if v.config.HighlightDiagnostics == nil || !*v.config.HighlightDiagnostics {
 			// HighlightDiagnostics is now not on - remove existing text properties
-			v.redefineHighlights([]types.Diagnostic{}, true)
+			v.BatchStart()
+			for bufnr, buf := range v.buffers {
+				if !buf.Loaded {
+					continue // vim removes properties when a buffer is unloaded
+				}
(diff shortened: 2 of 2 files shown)


cmd/govim/highlight.go
```diff
@@ -51,7 +51,7 @@ func (v *vimstate) textpropDefine() error {
 }
 
 func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) error {
-	if !force && (v.config.HighlightDiagnostics == nil || !*v.config.HighlightDiagnostics) {
+	if v.config.HighlightDiagnostics == nil || !*v.config.HighlightDiagnostics {
```
myitcv: It bothers me that `force` now means something different for quickfix, signs and highlights. Worth fixing now, or would you like to leave that for a later PR?

cmd/govim/vimstate.go
```diff
@@ -113,7 +113,18 @@ func (v *vimstate) setConfig(args ...json.RawMessage) (interface{}, error) {
 	if !vimconfig.EqualBool(v.config.HighlightDiagnostics, preConfig.HighlightDiagnostics) {
 		if v.config.HighlightDiagnostics == nil || !*v.config.HighlightDiagnostics {
 			// HighlightDiagnostics is now not on - remove existing text properties
-			v.redefineHighlights([]types.Diagnostic{}, true)
+			v.BatchStart()
+			for bufnr, buf := range v.buffers {
```
myitcv: It also bothers me that we leak this sort of detail into the handling of config changes. Worth fixing now, or would you like to leave that for a later PR?

leitzler: That's a good point. I did consider breaking out the feature to a `removeHighlights` but went with this since it is how it's done for both signs and quickfix entries above.

cmd/govim/highlight.go
leitzler: As per our slack conversation, I'll leave it as is for now.

When this is merged we'll have the following meaning of `force`:
- Highlights: Update even if there are no new diagnostics
- Signs: Update even if there are no new diagnostics
- Quickfix: Update even if there are no new diagnostics AND ignore if auto quickfix is disabled via config

PR #734, opened 2020-01-28
ci: enable gopls verbose logging for CI runs

myitcv:


myitcv:
Merging as this isn't really very interesting.

PR #737, opened 2020-01-28
cmd/govim: extract text property removal to own func

leitzler:
This change extracts text property removal to it's own func.

It acts as gardening for implementing references highlights where we'll
start using another text property ID, and therefore it is added as an
internal type.

### DIFF
--- a/cmd/govim/highlight.go
+++ b/cmd/govim/highlight.go
@@ -62,19 +62,10 @@ func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) erro
 		return nil
 	}
 
+	v.removeTextProps(types.DiagnosticTextPropID)
+
 	v.BatchStart()
 	defer v.BatchCancelIfNotEnded()
-	for bufnr, buf := range v.buffers {
-		if !buf.Loaded {
-			continue // vim removes properties when a buffer is unloaded
-		}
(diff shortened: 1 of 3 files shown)


cmd/govim/highlight.go
```diff
@@ -62,19 +62,10 @@ func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) erro
 		return nil
 	}
 
+	v.removeTextProps(types.DiagnosticTextPropID)
```
myitcv: What's slightly awkward is that we now have two batches. One way of dealing with this would be to optionally allow the caller (this method) to have started the batch. That would require us to have `BatchStartIfNeeded() bool` that returns whether the batch was started by that call or not. Then we could update `BatchCancelIfNotEnded` and `BatchEnd` to both take a bool that only does something if the argument is `true`.

leitzler: Did play around a bit with different options, but landed with `BatchStartIfNeeded() bool`, and then the caller is responsible of only call `BatchEnd()` / `BatchCancelIfNotEnded()` if the bool was true. It was the most clean solution that didn't add a lot of noise to other calls. We might want to revisit the batch mechanism in the future if we end up using `BatchStartIfNeeded()` a lot.

PR #741, opened 2020-01-29
cmd/govim: add reference highlighting when user goes idle

leitzler:
This change adds reference highlighting of the thing under the cursor,
when the user goes idle. As soon as the user goes idle, it will send a
documentHighlight request to gopls, and add a text property according to
the response.

To alter style/format, override the highlight group GOVIMReferences with
your prefered highlight setting.

### DIFF
--- a/cmd/govim/highlight.go
+++ b/cmd/govim/highlight.go
@@ -1,12 +1,16 @@
 package main
 
 import (
+	"context"
 	"fmt"
 
+	"github.com/govim/govim"
 	"github.com/govim/govim/cmd/govim/config"
+	"github.com/govim/govim/cmd/govim/internal/golang_org_x_tools/lsp/protocol"
 	"github.com/govim/govim/cmd/govim/internal/types"
 )
 
+// propDict is the representation of arguments used in vim's prop_type_add()
 type propDict struct {
 	Highlight string `json:"highlight"`
 	Combine   bool   `json:"combine,omitempty"`
@@ -14,6 +18,22 @@ type propDict struct {
 	StartIncl bool   `json:"start_incl,omitempty"`
 	EndIncl   bool   `json:"end_incl,omitempty"`
 }
+
+// propAddDict is the representation of arguments used in vim's prop_add()
+type propAddDict struct {
+	Type    string `json:"type"`
+	ID      int    `json:"id"`
+	EndLine int    `json:"end_lnum"`
+	EndCol  int    `json:"end_col"` // Column just after the text
+	BufNr   int    `json:"bufnr"`
+}
+
+// assertPropAdd is used when we add text properties that might fail due to the fact
+// that the buffer might have changed since the text properties was calculated.
+// There are two vim errors that we like to suppress, invalid line and invalid column.
+var assertPropAdd AssertExpr = AssertIsErrorOrNil(
+	"^Vim(let):E964:", // Invalid column (col) passed to vim prop_add()
+	"^Vim(let):E966:") // Invalid line (lnum) passed to vim prop_add()
 
 func (v *vimstate) textpropDefine() error {
 	v.BatchStart()
@@ -38,6 +58,12 @@ func (v *vimstate) textpropDefine() error {
 	v.BatchChannelCall("prop_type_add", config.HighlightHoverDiagSrc, propDict{
 		Highlight: string(config.HighlightHoverDiagSrc),
 		Combine:   true, // Combine with syntax highlight
+		Priority:  types.SeverityPriority[types.SeverityErr] + 1,
+	})
+
+	v.BatchChannelCall("prop_type_add", config.HighlightReferences, propDict{
+		Highlight: string(config.HighlightReferences),
+		Combine:   true,
 		Priority:  types.SeverityPriority[types.SeverityErr] + 1,
 	})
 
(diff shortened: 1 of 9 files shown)


cmd/govim/vimstate.go
```diff
@@ -175,9 +164,15 @@ func (v *vimstate) setUserBusy(args ...json.RawMessage) (interface{}, error) {
 	var isBusy int
 	v.Parse(args[0], &isBusy)
 	v.userBusy = isBusy != 0
+
+	if err := v.updateReferenceHighlight(!v.userBusy); err != nil {
```
myitcv: As discussed on Slack, I think we need to make this call async. We need to ensure there is close to zero latency when moving from "not busy" to "busy". The tail latency of a call to `gopls` will kill this.

leitzler:
Note to self - this depends on https://go-review.googlesource.com/c/tools/+/215742/ being merged into gopls. It includes a fix for request cancellation.

myitcv:
Just tried this locally.

I don't think we should be highlighting the identifier under the cursor - that looks weird and wasn't what I was expecting. Especially when there is just a single use in view; you lose the cursor.

I wonder too whether we should do something different from the `invert` highlighting. To my eye it's too "bold". Perhaps it's a good default, but is there a way we can "50% invert" or similar?

cmd/govim/highlight.go
```diff
@@ -86,18 +105,116 @@ func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) erro
...
+			return nil
+		}
+		v.currentReference = nil
+		v.removeTextProps(types.ReferencesTextPropID)
+		return nil
+	}
+
+	// Cancel any ongoing requests to make sure that we only process the
+	// latest response.
+	if v.cancelDocHighlight != nil {
```
myitcv: Access to `cancelDocHighlight` is racey. Needs to be guarded by a mutex.

cmd/govim/highlight.go
```diff
@@ -86,18 +105,116 @@ func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) erro
...
+		v.cancelDocHighlight()
+		v.cancelDocHighlight = nil
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	v.cancelDocHighlight = cancel
+
+	v.tomb.Go(func() error {
+		v.redefineReferenceHighlight(ctx, b, pos)
+		cancel()
+		v.cancelDocHighlight = nil
```
myitcv: Races here.

cmd/govim/highlight.go
```diff
@@ -86,18 +105,116 @@ func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) erro
...
+	if v.cancelDocHighlight != nil {
+		v.cancelDocHighlight()
+		v.cancelDocHighlight = nil
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	v.cancelDocHighlight = cancel
+
+	v.tomb.Go(func() error {
+		v.redefineReferenceHighlight(ctx, b, pos)
+		cancel()
```
myitcv: For the sake of consistency with the code and comment above, this should probably be:

```
v.cancelDocHighlight()
```

leitzler:
> I don't think we should be highlighting the identifier under the cursor

Interesting idea, I think there are two things that may impact UX with this. Highlighting a single identifier acts as visual confirmation that highlighting has been performed, and that the identifier is indeed the only one (if, for example, you have a very long updatetime). But I lean towards not highlight single hits anyway to make the feature less.. chatty/noisy.

The second thing that might impact UX is that I noticed that `gopls` is surprisingly powerful when it comes to highlighting and also highlights, for example, return statements.
E.g. this `main.go`:
```go
package main

import "fmt"

func main() { 
    foo()
}

func foo() error { 
    // lots of code
    return fmt.Errorf("failed")
}
```
Placing the cursor at the `m` in `return fmt.Errorf("failed")` will highlight both the fmt import (under the cursor + the import line) and the return statement (entire `fmt.Errorf...`) as well as `error` in the func signature.

If we remove the two highlights under the cursor, we will end up highlighting:
* `"fmt"` in `import "fmt"`
* `error` in `func foo() error {`

Would that be a behavior in lines with what you expect?

leitzler:
> I wonder too whether we should do something different from the invert highlighting. To my eye it's too "bold". Perhaps it's a good default, but is there a way we can "50% invert" or similar?

Totally agree. Just to add a bit of context to why I picked `invert` as suggestion of default. We could use something like bold and/or italic, but when I tried it out it was too subtle. A better way would be to change the background color, but since users might have a huge variation of themes, the background color we pick would stand out way too much for some users.

That leaves us with foreground color. A bit better, but could still conflict with different themes (and the fact that it is hard to pick a color that works for both light and dark backgrounds :)).

A "slightly inverted" would be the best, but I haven't found a way to apply something like that. And I don't think a feature suggestion to add opacity to `inverse` in vim would be long lived ;)

myitcv:
> A "slightly inverted" would be the best, but I haven't found a way to apply something like that. And I don't think a feature suggestion to add opacity to `inverse` in vim would be long lived ;)

No, that background makes total sense. I was just thinking out loud.

cmd/govim/highlight.go
```diff
@@ -86,18 +105,116 @@ func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) erro
...
+			TextDocumentPositionParams: protocol.TextDocumentPositionParams{
+				TextDocument: protocol.TextDocumentIdentifier{
+					URI: string(b.URI()),
+				},
+				Position: cursorPos.ToPosition(),
+			},
+		},
+	)
+
+	if err == context.Canceled {
```
myitcv: Should this be something like:

```go
select {
case <-ctx.Done():
    return
default:
}
if err != nil {
    v.Logf("documentHighlight call failed: %v", err)
    return
}
```

cmd/govim/highlight.go
```diff
@@ -86,18 +105,116 @@ func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) erro
...
+		// A newer request canceled this one, so we shouldn't process the results here
+		return
+	} else if err != nil {
+		v.Logf("documentHighlight call failed: %v", err)
+		return
+	}
+
+	v.govimplugin.Schedule(func(govim.Govim) error {
+		return v.handleDocumentHighlight(b, cursorPos, res)
```
myitcv: Within this scheduled func we probably want another

```
select {
case <-ctx.Done():
    return
default:
}
```

cmd/govim/highlight.go
leitzler: Hah, removed the mutex and afterwards realized that I needed to cancel here to free up resources. Will add it back.

cmd/govim/highlight.go
```diff
@@ -41,6 +54,12 @@ func (v *vimstate) textpropDefine() error {
 		Priority:  types.SeverityPriority[types.SeverityErr] + 1,
 	})
 
+	v.BatchChannelCall("prop_type_add", config.HighlightReferences, propDict{
```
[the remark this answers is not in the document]
leitzler: No, doesn’t have to be here. I did put it here to have one single place where all text properties are defined.

Costs would be a another separate call, i.e. round trip to vim. The function is only called once so not a huge cost.

leitzler:
Currently blocked by https://github.com/golang/go/issues/37159

cmd/govim/highlight.go
leitzler: I don't think we should. Since I cancel the context to free up resources after the request to gopls, and the schedule execution might be scheduled far later, it might skip a valid execution.

I think that is what happened in the failed `gvim` tests (flakes when I had the select case). From the logs the only difference was a `Hover` request in the `gvim` logs, something I guess took enough time for us to hit this select-case return.

cmd/govim/highlight.go
```diff
@@ -86,18 +105,127 @@ func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) erro
...
+			v.cancelDocHighlight = nil
+		}
+		v.cancelDocHighlightLock.Unlock()
+		return nil
+	})
+
+	return nil
+}
+
+func (v *vimstate) redefineReferenceHighlight(ctx context.Context, b *types.Buffer, cursorPos types.Point) {
```
myitcv: This method should be defined on `func (g *govimplugin)`

cmd/govim/highlight.go
```diff
@@ -86,18 +105,127 @@ func (v *vimstate) redefineHighlights(diags []types.Diagnostic, force bool) erro
...
+	case <-ctx.Done():
+		return
+	default:
+	}
+	if err != nil {
+		v.Logf("documentHighlight call failed: %v", err)
+		return
+	}
+
+	v.govimplugin.Schedule(func(govim.Govim) error {
```
myitcv: This then becomes `g.Schedule`

cmd/govim/highlight.go
myitcv: Hmm, not sure. The reason I think you want to cancel here is that it's your last chance to do so before applying a potentially bad highlight. Because if you don't check the cancel here and the highlight you are about to apply has been cancelled, then it will look wrong until the the "correct" call returns.

Also I think you need to be re-checking whether the user is busy or not in here. Because if they are (which is the most likely cause of the current highlight having been cancelled) you definitely don't want to do anything.

cmd/govim/highlight.go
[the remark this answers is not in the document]
myitcv: Sounds good.

cmd/govim/highlight.go
leitzler: Yeah, I see the reasoning. I re-added the `ctx.Done()` check here, and added a busy check. I have a gut feeling that the busy check might cause unexpected behaviour, but can't seem to come up with such case.

cmd/govim/highlight.go
leitzler: Done

cmd/govim/highlight.go
leitzler: Done

myitcv:
I've actually just found an even better default on my system:

```
hi GOVIMReferences ctermbg=67 ctermfg=0 guibg=DarkGrey
```

Screen Shot 2020-02-16 at 10 11 59

I don't think this changes what we do though: let's get this merged and I'm sure people with a good artistic/UI/UX eye will tell us what's best 👍

leitzler:
They look really good! Yeah, let's start with `reverse`, and see if we can change to something better that suits both light and dark themes later.

cmd/govim/highlight.go
```diff
@@ -15,6 +19,22 @@ type propDict struct {
 	EndIncl   bool   `json:"end_incl,omitempty"`
 }
 
+// propAddDict is the representatino of arguments used in vim's prop_add()
```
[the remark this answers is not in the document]
leitzler: ops, fixed

cmd/govim/highlight.go
```diff
@@ -15,6 +19,22 @@ type propDict struct {
...
+type propAddDict struct {
```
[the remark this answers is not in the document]
leitzler: ack

PR #746, opened 2020-01-31
ci: first cut of running tests on macOS

myitcv:


myitcv:
@leitzler - I've updated this to not use a third-party GitHub action (per the "advice" in https://blog.myitcv.io/2020/02/04/portable-ci-cd-with-pure-go-github-actions.html). Instead we use a local pure Go action. Perhaps you could give it a look?

.github/workflows/test.yml
```diff
@@ -58,3 +58,31 @@ jobs:
...
+        go_version: ['1.14.x']
+        vim_version: ["master"]
+    runs-on: ${{ matrix.os }}
+    env:
+      VIM_FLAVOR: vim
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@722adc63f1aa60a57ec37892e133b1d319cae598
+    - name: Install Vim
+      uses: ./github/actions/setupvim
```
leitzler: Is the idea to keep the action as a part of the `govim/govim` repo?
Reason for asking is that the `README.md` header points to another (non-existing/private) repo, and the `go.mod` have a module name that is close to real but doesn't exist either.

github/actions/setupvim/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/govim/govim/githubactions/github/actions/setupvim
```
leitzler: This line, I guess it doesn't matter that much since it isn't imported by anyone. But maybe it's clearer to use something without host to avoid confusion? Like just `setupvim`.

github/actions/setupvim/README.md
```diff
@@ -0,0 +1,3 @@
+### `github.com/myitcv/action-setup-vim`
```
leitzler: This line

internal/cmd/genconfig/genconfig.go
```diff
@@ -289,6 +290,36 @@ jobs:
...
       if: failure()
       run: ./_scripts/afterFailure.sh
+
+  test-macos:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest]
+        go_version: ['1.13.x', '1.14.x']
+        vim_version: ["stable"]
```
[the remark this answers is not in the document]
myitcv: Comment superseded.

.github/workflows/test.yml
myitcv: > Is the idea to keep the action as a part of the govim/govim repo?

For now, yes. As you have keenly observed, it originated elsewhere as I was hacking around writing https://blog.myitcv.io/2020/02/04/portable-ci-cd-with-pure-go-github-actions.html. Doesn't really matter where it lives to be honest, because I don't suspect it will live for any reason other than our own use.

github/actions/setupvim/go.mod
myitcv: If we do that, this then becomes unusable by anyone other than us because the module path (`setupvim`) would not match the address path (`github.com/govim/govim/github/actions/setupvim`). But perhaps you were referring to my typo in the middle of the module path, in which case thanks for the spot :)

github/actions/setupvim/README.md
myitcv: Fixed to match the module path.

myitcv:
Thanks for taking a look.

> Future PR's should at least include artefact uploads.

Yes, that should be relatively straightforward using `_scripts/captureLogs.sh`

PR #748, opened 2020-01-31
cmd/govim: switch to the didChangeWatchedFiles API

### LINKED ISSUE #187: cmd/govim: add support for generated files changing
On the `gopls` side this is being tracked by https://github.com/golang/go/issues/31553

On Vim/`govim` side we are going to need to add some level of file watcher, because this is not something that Vim supports out of the box. Just to be explicit, the requirement here is that _any_ file considered as "relevant" by `gopls` should effectively be monitored by the LSP client (`govim`). Vim "monitors" files that it has open in buffers but doesn't do anything automatically (i.e. the user can choose to ignore the fact the file on disk has changed underneath the buffer).

So I think the requirement here is:

* `gopls` to tell `govim` the set of file globs it should watch
* Vim to be in "control" of the subset of files matching those globs that are open in buffers, i.e. `govim` to use the existing `didOpen`/`didChange` notification for those files, _not_ the file contents of disk
* `govim` to watch what remains and notify via `didChangeWatchedFiles`

findleyr:
Now that gopls supports the didChangeWatchedFiles notification, govim no
longer needs to synthetically open files on external file changes.

This change updates the fswatcher package, along with watcher.go, to use
didChangeWatchedFiles. Notably, this change requires tracking file
creation separately from modification or removal.

Tests are updated to exercise some of this new behavior.

Fixes #187

### DIFF
--- a/cmd/govim/testdata/scenario_default/complete_watched.txt
+++ b/cmd/govim/testdata/scenario_default/complete_watched.txt
@@ -1,12 +1,9 @@
 # Test that ominfunc complete works where the completion is made
 # available in a file that is not loaded via the editor.
-#
-# We add in a generous sleep to ensure the watch event has been
-# handled.
 
 vim ex 'e main.go'
 cp const.go.orig const.go
-errlogmatch '&protocol\.DidOpenTextDocumentParams\{\n\S+:\s+TextDocument: protocol.TextDocumentItem{URI:"file://'$WORK/const.go
+errlogmatch '&protocol\.DidChangeWatchedFilesParams\{\n\S+:\s+Changes: \{\n\S+:\s+\{URI:"file://'$WORK/const.go'", Type:1\}'
 vim ex 'call cursor(6,16)'
 vim ex 'call feedkeys(\"i\\<C-X>\\<C-O>\\<C-N>\\<C-N>\\<ESC>\", \"x\")'
 vim ex 'w'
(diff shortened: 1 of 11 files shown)


cmd/govim/testdata/scenario_default/complete_watched.txt
```diff
@@ -1,20 +1,16 @@
...
+errlogmatch '&protocol\.DidChangeWatchedFilesParams\{\n\S+:\s+Changes: \{\n\S+:\s+\{URI:"file://'$WORK/const.go'", Type:1\}'
 vim ex 'call cursor(6,16)'
 vim ex 'call feedkeys(\"i\\<C-X>\\<C-O>\\<C-N>\\<C-N>\\<ESC>\", \"x\")'
 vim ex 'w'
 cmp main.go main.go.golden
 
 # Assert that we have received no error (Type: 1) or warning (Type: 2) log messages
-# Disabled pending resolution to https://github.com/golang/go/issues/34103
-# errlogmatch -start -count=0 'LogMessage callback: &protocol\.LogMessageParams\{Type:(1|2), Message:".*'
+errlogmatch -start -count=0 'LogMessage callback: &protocol\.LogMessageParams\{Type:(1|2), Message:".*'
```
findleyr: Huh, this was passing for me, and that issue was resolved, so I uncommented. Perhaps this needs to be switched to use vimexprwait.

PR #750, opened 2020-02-01
cmd/govim: really ensure go-to-def tests have stable go.mod files

myitcv:
In e5653c8e we sort to make go-to-def testscript tests go.mod stable. We
missed an instance as part of that change. Fix that.


myitcv:
Merging as this is really an uninteresting change.

PR #754, opened 2020-02-03
cmd/govim: fix embarrassingly bad bug in completion textedit handling

myitcv:
As part of the current completion code that handles applying of
additional text edits we have:

	for _, c := range v.lastCompleteResults.Items {
		if c.Label == chosen.Abbr {
			match = &c
		}
	}

In a previous change, compltion items changes from being a pointer to a
non-pointer value. Hence we flipped the setting of match from c to &c.
But we forgot the break (which arguably could/should have been there
before). Which means that a further iteration of the for loop after a
match will cause the match to find the text edits from another match.
Specifically, we will be left with the additional text edits from the
last match in the list.

Fix that by breaking out of the loop on a match.

As part of this also fix a test which was incorrectly updated as a
result of this bug (which actually makes the bug itself even worse!)

### DIFF
--- a/cmd/govim/complete.go
+++ b/cmd/govim/complete.go
@@ -81,6 +81,7 @@ func (v *vimstate) completeDone(args ...json.RawMessage) error {
 	for _, c := range v.lastCompleteResults.Items {
 		if c.Label == chosen.Abbr {
 			match = &c
+			break
 		}
 	}
 	if match == nil {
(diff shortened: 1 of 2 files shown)


myitcv:
Incidentally, the reason we appear not to have consistently fallen foul of this is that the second completion result in the test changed as part of this commit is from unimported completions. It would appear that the test generally completes so quickly that unimported completion candidates are not yet available, hence we only get one result most of the time. But on rare occasions we do get two results, the second of which is the `log` package. Hence the failure seen in https://github.com/govim/govim/runs/423494363?check_suite_focus=true

PR #756, opened 2020-02-04
ci: add branchCheck to postRun script

leitzler:
Jobs that are skipped in WIP branches have no $ARTEFACTS directory
so those will fail during post run.

This change adds a branchCheck to exit gracefully.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,7 @@ jobs:
     - name: Tidy up
       run: ./_scripts/postRun.sh
     - name: Upload artefacts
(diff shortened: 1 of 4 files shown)


myitcv:
I think we can probably enhance this somewhat to then skip the uploading of artefacts, but good for now.

leitzler:
> I think we can probably enhance this somewhat to then skip the uploading of artefacts, but good for now.

Yeah, upload step failed too so I added an exported ENV var.  We should probably tidy up the process a bit more, but it would be nice to get things working.

PR #757, opened 2020-02-05
deps: upgrade x/tools and gopls to 6e8b36d2

myitcv:


myitcv:
Should fix the CRON build failures we are seeing. FYI @findleyr

PR #758, opened 2020-02-05
ci: upgrade to go1.14rc1 and Vim v8.2.0213 in the build matrix

myitcv:


testsetup/testsetup.go
```diff
@@ -73,7 +73,7 @@ func (c Command) BuildCommand(args ...string) *exec.Cmd {
 }
 
 var (
-	GoVersions = []string{"go1.12.14", "go1.13.5", "go1.14beta1"}
+	GoVersions = []string{"go1.12.14", "go1.13.5", "go1.14rc1"}
```
leitzler: Shouldn't we update to `go1.12.16` and `go1.13.7` while at it?

myitcv: Yes, we should. Nice spot :)

PR #764, opened 2020-02-10
ci: Ensure that artifacts are created and not overwritten

leitzler:
Artifacts are only written when a job pass, and use the
same name for all jobs so they overwrite each other.

This change ensures that artifacts are uploaded on both
success and failure, and that the artifact zip is named
based on which job it came from.


myitcv:
Per Slack, on hold pending an answer on if/why the condition changes are needed.

internal/cmd/genconfig/genconfig.go
```diff
@@ -295,13 +295,14 @@ jobs:
       if: success()
       run: ./_scripts/runDockerRun.sh
     - name: Tidy up
+      if: success() || failure()
       run: ./_scripts/postRun.sh
     - name: Upload artefacts
-      if: success() && env.CI_SKIP_JOB != 'true'
+      if: (success() || failure()) && env.CI_UPLOAD_ARTIFACTS == 'true' && !(env.CI_SKIP_JOB == 'true')
```
myitcv: I'm not sure this is right. We only want to run this step if the previous step succeeded. Why is the previous logic here not right? Because the addition of the env var means that we will only ever run this step if the previous step ran to completion, and hence success.

myitcv: Per Slack, you are 100% right. My apologies. What's definitely changed here is something on the GitHub side.

PR #769, opened 2020-02-12
ci: remove errant RACE_BUILD env var in workflow config

myitcv:


myitcv:
Mac builds being slow again...

PR #770, opened 2020-02-12
cmd/govim: refactor batch call/expr handling

myitcv:


myitcv:
I added a `ch_log` statement. Easy to do; good catch. Thanks

PR #773, opened 2020-02-14
deps: upgrade x/tools and gopls to 88be0131

myitcv:
* internal/jsonrpc2: fix races in cancellation 88be0131
* internal/lsp/lsprpc: add an LSP forwarder and regtest environment 0fd2d649
* internal/lsp: add codelens for go.mod dependency upgrades 98b3097d


myitcv:
No `govim` changes required here. Build is green (modulo slow, flakey Mac build). Merging

PR #774, opened 2020-02-14
cmd/govim: define GOVIMWorkspaceSymbol to expose gopls' workspace/symbol

myitcv:
This allows calls like:

    call GOVIMWorkspaceSymbol({'query': 'main'})

### DIFF
--- a/cmd/govim/testdata/scenario_default/workspace_symbol.txt
+++ b/cmd/govim/testdata/scenario_default/workspace_symbol.txt
@@ -1,0 +1,39 @@
+# Test that GOVIMWorkspaceSymbol works as expected
+
+vim -indent expr 'GOVIMWorkspaceSymbol({\"query\": \"main\"})'
+cmpenv stdout symbols.golden
+! stderr .+
+
+# Assert that we have received no error (Type: 1) or warning (Type: 2) log messages
+# Disabled pending resolution to https://github.com/golang/go/issues/34103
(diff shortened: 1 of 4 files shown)


cmd/govim/testdata/scenario_default/workspace_symbol.txt
```diff
@@ -0,0 +1,39 @@
+# Test that GOVIMWorkspaceSymbol works as expected
+
+vim -indent expr 'GOVIMWorkspaceSymbol({\"query\": \"main\"})'
```
findleyr: Is govim's philosophy generally to pass through the request parameters like this? I think it would be nicer to have GOVIMWorkspaceSymbol accept a string.

myitcv: > Is govim's philosophy...

I don't think I/we have ever had a philosophy ;-)

Joking aside, this is the first `gopls`/LSP function we have exposed directly to Vim with no intermediate step. Part of the reason for simply acting as a pass-through is that otherwise we need to establish a process for flattening the params struct:

https://github.com/golang/tools/blob/ea829e2eb296fe30ae77de319c26e88e3b165652/internal/lsp/protocol/tsprotocol.go#L3525-L3533

working out what types to use (for example `WorkDoneToken` would likely need to be a `*interface{}` to avoid the zero-value for a `string`/`int` meaning something) and then documenting those parameters. 

So in many respects I was being lazy.

Ideas very much welcomed for alternative approaches.

myitcv:
This is no longer required AFAICT. And it's 4 years old.

PR #775, opened 2020-02-14
deps: upgrade x/tools and gopls to c2296495

myitcv:
* internal/lsp/cache: fix crash fixing curlies near EOF c2296495


myitcv:
No `govim` changes so merged.

PR #789, opened 2020-02-21
deps: upgrade x/tools and gopls to 57f3fb51

leitzler:
* gopls: update Staticcheck to 2020.1.2 57f3fb51
* internal/lsp: create more descriptive temp go.mod name 9c32af92
* internal/lsp/regtest: skip flaky TestGoToStdlibDefinition fe62aff3
* internal/lsp/source: improve completions at file scope 8a925fa4
* internal/lsp/regtest: increase test timeout to 60s 947cbf19


myitcv:
Just need to run the revendor script

leitzler:
:+1: seems like there is a bug when running that script under macOS. I'll take a look later why it missed those files.

PR #792, opened 2020-02-22
cmd/govim: switch ftdetect for go.mod files to be *.mod

myitcv:
In Go 1.14 the -modfile flag lands for cmd/go. This allows an alternate
go.mod file to be provided for reads/writes during module resolution.
The alternate file must have a .mod extension, hence we should update
our file detection logic to be for *.mod.


myitcv:
@cespare - care to give this a glance over?

cespare:
Seems fine to me.

Issue #687, opened 2020-01-18
cmd/govim: add tests to verify that turning off unimported completions works


myitcv:
Now that https://github.com/govim/govim/commit/d80f0e21c2e3291f93bd60d3d70079f877834d85 has landed with unimported completions turned on by default, we need tests to verify that turning that feature off works.

tjcain:
Hi @myitcv, I would like to take a stab at this if that is ok?

Would it make sense to give this case its own scenario i.e. `govim/cmd/govim/testdata/scenario_completeunimported_off` or make it a case in the existing `scenario_completeunimported`?

myitcv:
Hi @tjcain 

> I would like to take a stab at this if that is ok?

Absolutely fine. 

I've just submitted #793 to complete a bit of house work. So once that is merged please do have a stab at this. https://github.com/govim/govim/wiki/govim-tests should help provide a bit of background, but if you do have any questions please do feel free to ask here or in `#govim-dev` on [Gophers Slack](https://invite.slack.golangbridge.org/)

PR #796, opened 2020-02-24
cmd/govim: add tests to verify turning off unimported completions works (#687)

tjcain:
Following d80f0e2 unimported completions are turned on by default. This
commit adds tests to ensure setting CompleteUnimported to false disables
completion for unimported packages.

### DIFF
--- a/cmd/govim/testdata/scenario_no_completeunimported/no_complete.txt
+++ b/cmd/govim/testdata/scenario_no_completeunimported/no_complete.txt
@@ -1,0 +1,41 @@
+# Test that completing of unimported std library packages is disabled
+# when CompleteUnimported=0
+
+vim ex 'e main.go'
+
+# Attempt unimported completion; we should get zero results.
+# If we did get results (from fmt) then attempting completion
+# would expand from fmt.Pr to fmt.Print, i.e. the longest match
+# of the returned candidates
+vim ex 'call cursor(4,1)'
+vim normal Sfmt.Pr
+vim ex 'execute \"normal A\\<C-X>\\<C-O>(\\\"Hello\\\")\"'
+
+vim ex 'noau w'
+exec cat main.go
+vim ex message
+
+# Check import has not been added
+vim ex 'noau w'
+cmp main.go main.go1.golden
+
+# Assert that we have received no error (Type: 1) or warning (Type: 2) log messages
+# Disabled pending resolution to https://github.com/golang/go/issues/34103
+# errlogmatch -start -count=0 'LogMessage callback: &protocol\.LogMessageParams\{Type:(1|2), Message:".*'
+
+-- go.mod --
+module mod.com
+
+go 1.12
+-- main.go --
+package main
+
+func main() {
+	fmt.Pr
+}
+-- main.go1.golden --
+package main
+
+func main() {
+	fmt.Pr("Hello")
+}
--- a/cmd/govim/testdata/scenario_no_completeunimported/no_complete_empty_file.txt
+++ b/cmd/govim/testdata/scenario_no_completeunimported/no_complete_empty_file.txt
@@ -1,0 +1,42 @@
+# Test that completing of unimported std library packages is disabled
+# when files on disk are initially empty and CompleteUnimported=0
+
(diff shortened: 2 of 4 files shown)


cmd/govim/testdata/scenario_no_completeunimported/no_complete.txt
```diff
@@ -0,0 +1,42 @@
...
+-- go.mod --
+module mod.com
+
+go 1.12
+-- main.go --
+package main
+
+func main() {
+	fmt.Pr
+
```
myitcv: Any reason for the blank lines here?

cmd/govim/testdata/scenario_no_completeunimported/no_complete.txt
```diff
@@ -0,0 +1,42 @@
+# Test that completing of unimported std library packages is disabled
+# when CompleteUnimported=0
+
+vim ex 'e main.go'
+
+# Attempt unimported completion
+vim ex 'call cursor(4,1)'
+vim normal Sfmt.Pr
+vim ex 'execute \"normal A\\<C-X>\\<C-O>\\<C-N>\\<C-N>(\\\"Hello\\\")\"'
```
myitcv: As I mentioned on Slack, given we are testing in the context of the `fmt` package which has three functions that share a `Print` prefix and the behaviour of omni-completion, the `<C-n><C-n>` keystrokes are not strictly required. i.e. there are no candidates returned, so firing the keystrokes will be a no-op. 

So perhaps we can remove them in all tests you're adding to remove an potential confusion with the reader thinking we are cycling through results? 

Indeed, might just be worth amending the comment above to something like:

```
# Attempt unimported completion; we should get zero results.
# If we did get results (from fmt) then attempting completion
# would expand from fmt.Pr to fmt.Print, i.e. the longest match
# of the returned candidates
```

cmd/govim/testdata/scenario_no_completeunimported/no_complete_empty_file.txt
```diff
@@ -0,0 +1,39 @@
+# Test that completing of unimported std library packages is disabled
+# when files on disk are initially empty and CompleteUnimported=0
+
+# Setup new buffer and verify contents are as expected
+vim ex 'e main.go'
+vim ex 'r main.go.orig | 0d_'
+vim ex 'noau w! check'
+cmp check main.go.orig
+
+# Attempt unimported completion
```
myitcv: Might be worth copying the comment from the first test here too

cmd/govim/testdata/scenario_no_completeunimported/no_complete.txt
tjcain: No good reason, an artefact from copying tests from the default complete case, I will remove them.

myitcv:
@tjcain - sorry, one thing I've just spotted (reviewing the commit message is not, sadly, part of GitHub's flow here, unlike Gerrit).

The `(#xxx)` suffix is added to commits automatically by GitHub because we are using "squash and merge". But despite using that feature per [the wiki](https://github.com/govim/govim/wiki/Contributors) we rely on the commit message being "good", otherwise if left unsquashed the person doing the merge has to do the hard work of coming up with a good message!

Your message is great, with the exception of some really small nits:

* the superfluous suffix on the first line - this will simply be confusing when the PR is merged because it refers to the issue you are fixing. It will become the PR number itself so the linear history of `govim` can be quickly and easily traced back to the PRs that created commits
* the first line is a bit verbose
* the message is missing the "Fixes" directive

Therefore, how about something like?


```
cmd/govim: verify turning off unimported completions works

Following d80f0e2 unimported completions are turned on by default. This
commit adds tests to ensure setting CompleteUnimported to false disables
completion for unimported packages.

Fixes #687
```

That way #687 will be closed automatically when we merged.

Thanks

tjcain:
> @tjcain - sorry, one thing I've just spotted (reviewing the commit message is not, sadly, part of GitHub's flow here, unlike Gerrit).

I will get that fixed up now, thank you for the great feedback.

myitcv:
Will ignore the macOS flake for now

PR #797, opened 2020-02-25
cmd/govim: fix embarrassingly bad bug in buffer listener reset

myitcv:
We were only setting a buffer's listener to 0 (i.e. no listener) on
unload for buffers that were of interest to gopls. Hence for non-gopls
buffers we were getting strange lifecycle errors.

Add a test to verify that moving around between non gopls buffers works,
a test that failed before this fix.


myitcv:
Moved test away from using help to regular files.

PR #813, opened 2020-03-05
[WIP] deps: up required version of Vim/GVim to v8.1.2056

myitcv:


myitcv:
This has been long superseded. We now only test against a ~latest version of vim/gvim in any case.

PR #811, opened 2020-03-05
cmd/govim: fix handling of changed diagnostics from gopls

myitcv:

PR #818, opened 2020-03-09
cmd/govim: add buffer reload sign placment test

leitzler:
During review of #811 I noticed that we are lacking tests of sign
placment.

This change adds a test that ensures that signs are placed when loading
a unloaded buffer, after being temporary removed by a masking error.


myitcv:
Thanks @leitzler - I've now added this test as part of #811

PR #826, opened 2020-03-15
cmd/govim: add autoread loaded buffers support

leitzler:
If a file is updated on disk while it is open in vim, govim/gopls won't
pick up the change automatically.

This change adds a config setting "AutoreadLoadedBuffers" that is used
to reload buffers for files that change outside vim. It requires the
buffer to be "loaded" since it uses "checktime" to reload the file.

This feature is not enabled by default. Add the following to your vimrc
to enable it:

set autoread
set hidden
call govim#config#Set("AutoreadLoadedBuffers", 1)

"set autoread" will suppress a confirmation prompt when a buffer that
isn't modified is reloaded (a prompt will still be shown if the buffer
is modified). This line is optional.

"set hidden" will configure vim to keep buffers loaded when switched,
instead of the default that is to abandon them. This is required for the
feature to work with buffers that aren't switched away.

### DIFF
--- a/autoload/govim/config.vim
+++ b/autoload/govim/config.vim
@@ -120,6 +120,10 @@ function! s:validGoplsEnv(v)
   return [v:true, ""]
 endfunction
 
+function! s:validExperimentalAutoreadLoadedBuffers(v)
+  return s:validBool(a:v)
+endfunction
+
 function! s:validExperimentalMouseTriggeredHoverPopupOptions(v)
   if has_key(a:v, "line")
     if type(a:v["line"]) != 0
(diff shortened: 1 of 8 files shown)


myitcv:
Interesting. So you're suggesting this because approaches like https://unix.stackexchange.com/questions/149209/refresh-changed-content-of-file-opened-in-vim/383044#383044 are less reliable?

leitzler:
I think they are quite reliable, but doesn’t really cover all cases. I haven’t dived into how running `checktime` Each CursorHold affects modern SSDs (not at all?), but thought that since we already have the information via watchers, why not use it as a way to reduce calls a lot. The PR also only calls `checktime` for the specific buffer that gets a file watcher event.

The suggestion about FocusGained & BufEnter doesn’t seem to work for terminal vim (at least not mine).

To try it out, create two files in a package:
```
— main.go —
package main

func main() {
    _ = foo
}
— other.go —
// const foo = 0
```

Then open both in one vim session (with `main.go` visible/loaded). And have another vim instance with `other.go` open where you remove the comment and save the file.

With this PR, the first vim will pick up the change all the way with diags and all. (Also requires `set autoread` & `set hidden` in vimrc).

myitcv:
Per our exchange on Slack, probably best to mark this as "experimental" for now whilst we iterate and land on the "best" approach.

PR #829, opened 2020-03-18
govim: move plugin errorchan read to avoid data race

leitzler:
Embarrassing enough PR 828 fixed the deadlock but introduced a data
race where the plugin error channel was read in a goroutine created
before the channel was created.

This change fixes the data race by moving the reading goroutine to be
started after the channel creation.

### DIFF
--- a/govim.go
+++ b/govim.go
@@ -312,12 +312,13 @@ func (g *govimImpl) load() error {
 		g.Logf("No build info available")
 	}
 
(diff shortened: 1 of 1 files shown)


myitcv:
Please can you explain in the commit message where the data race is?

leitzler:
Sure, I updated the commit & PR description!

myitcv:
Thanks!

PR #832, opened 2020-03-24
deps: update x/tools and gopls to 09944062

myitcv:


myitcv:
Blocked on https://github.com/golang/go/issues/38042

findleyr:
To save you time debugging TestScripts/scenario_default/signs_unload_load_buffer, once you update gopls again...
https://storage.googleapis.com/golang-gopls_integration_tests/log-eaa8d2ea-7f0a-4a28-9a09-a0ab4d594f8e.txt

There is now a new (at first glance duplicate) diagnostic coming from https://golang.org/cl/222237. Perhaps govim should de-dupe it.

/cc @stamblerre

findleyr:
Regarding the blockage: you should now be able to work around golang/go#38042 by disabling telemetry: set GOVIM_GOPLS_FLAGS="-telemetry.disable" in your test environment.

Sorry for the hack, but @ianthehat is rewriting large chunks of the telemetry stack, so this blunt fix avoids trying to hit a moving target.

leitzler:
Just a thought, wouldn’t it make more sense to disable it by default with a flag to enable? Or are those panics strongly related to running in CI only?

myitcv:
> There is now a new (at first glance duplicate) diagnostic coming from https://golang.org/cl/222237. Perhaps govim should de-dupe it.

Per the previous instance of this, the fix currently lives in `gopls`: https://github.com/golang/go/issues/34494

> set GOVIM_GOPLS_FLAGS="-telemetry.disable" in your test environment.

We can do this, but I agree with @leitzler - telemetry should probably be off by default it it's not "ready".

myitcv:
> There is now a new (at first glance duplicate) diagnostic coming from https://golang.org/cl/222237. Perhaps govim should de-dupe it.

Raised as https://github.com/golang/go/issues/38136

myitcv:
This is currently blocked on the two issues discussed in https://github.com/govim/govim/pull/832#issuecomment-605499712

myitcv:
I've added a skip for the test that is failing pending a fix for https://golang.org/issues/38136

PR #835, opened 2020-03-29
cmd/govim: trigger buffer events for *.mod files

myitcv:
For completeness we now trigger buffer events for *.mod files.  It
remains highly discouraged to edit go.mod files by hand (because the go
command, and by extension gopls in the default govim configuration edits
it automatically).


myitcv:
This got done elsewhere along the way.

PR #844, opened 2020-04-06
cmd/govim: remove testdata directories from golang.org/x/tools "vendor"

myitcv:
These are useless so let's just remove them.


myitcv:
This does not touch any `govim` code. Merging

PR #845, opened 2020-04-06
deps: update x/tools and gopls to 44c82bac18ac

myitcv:


myitcv:
Only small tweaks to jsonrpc2 handlers required. No change to `govim` code. Merging when the build goes green.

PR #847, opened 2020-04-06
cmd/govim: tidy-up refactor

myitcv:


cmd/govim/flag.go
```diff
@@ -0,0 +1,51 @@
...
+
+func main() { os.Exit(main1()) }
+
+func main1() int {
+	err := mainerr()
+	if err == nil {
+		return 0
+	}
+	switch err := err.(type) {
+	case usageErr:
```
leitzler, line 42: This type should be returned from `mainerr(...)` right? Usage isn't printed as it is now.

cmd/govim/flag.go
```diff
@@ -0,0 +1,51 @@
...
+		fmt.Fprintln(os.Stderr, err)
+		flagSet.Usage()
+		return 2
+	case flagErr:
```
leitzler, line 46: This neither, is it just preparations for future PRs?

myitcv: Yes - just prep for a later PR. But your comments did help me spot a couple of fixes... which I've made.

PR #848, opened 2020-04-06
cmd/govim: first cut of command-line mode to query "parent"

### LINKED ISSUE #795: cmd/govim: command-line mode to query "parent" cmd/govim instance
A [recent exchange on Slack](https://gophers.slack.com/archives/CL8JDP1GD/p1581693138070200) discussed how to support the [LSP Symbol method](https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#workspace_symbol), especially in combination with (fuzzy) finders like [`fzf`](https://github.com/junegunn/fzf). Using `fzf` as a popular example of a (fuzzy) finder, we explored what would be required to use `fzf` in conjunction with the Symbol method. 

The Symbol method takes a query parameter and returns the set of matching results. `fzf` supports using fuzzy finding based on some input, selecting the desired result from the list of results, then doing something with that selection, e.g. opening a file (in a (v)split)

The slightly wrinkle in this case is that we need to re-query the Symbol method as the user types in the fuzzy finder. We raised this in https://github.com/junegunn/fzf/issues/1883

As the response indicates, an approach similar to https://github.com/junegunn/fzf.vim/issues/948 is required here because `fzf` cannot call back into Vim in order to re-call the Symbol method. 

Hence this issue is a placeholder for making changes in `govim` to support a "child" `govim` instance querying a "parent" `govim` instance to, at least, call the Symbol method by some means. Gathering some of the notes from the Slack exchange:

* Offer a means by which `govim` can be run as a cmd to connect to an existing `govim` process and run certain queries/commands etc
* In `github.com/govim/govim/testdriver` we already provide a means by which we expose, via a command that connects to `govim`, a means to arbitrarily run Vim commands, evaluate expressions etc. This command is the special `vim` command that is used in the `testscript` tests
* A decent first step would be to refactor the `vim` command functionality in `testdriver` to be a part of `cmd/govim`, enabled by a special flag (e.g. `-parent SOCKET`). Hence a call to the Symbol method would look like:

```
govim -parent :4567 vim expr "GOVIMSymbol({\"query\": \"main\"})"
```

* We could then even bypass Vim and have `govim` directly relay certain calls to `gopls`, e.g.

```
govim -parent :4567 lsp Symbol "({\"query\": \"main\"})"
```

This would cut out the overhead in going `govim (command) -> govim -> Vim -> govim -> gopls` and back again, and instead make the sequence `govim (command) -> govim -> gopls` and back again; the hops via Vim are not required at all here

### Command API

`govim` would gain a new flag, `-parent`, which would take the Unix/TCP/IP socket to connect to. This flag would indicate the caller is the child, and expect commands 

At a bare minimum we need to expose the Symbol method, and it probably makes most sense to make that a direct query against `gopls`, not via Vim. 

Therefore it probably makes sense to create an `lsp` subcommand which takes a method name and a JSON encoded argument:

```
govim -parent :4567 lsp Symbol "({\"query\": \"main\"})"
```

We could also expose a `vim` subcommand in a similar way to the `testdriver` `vim` command used in `testscript` tests:

```
govim -parent :4567 vim expr getqflist()
```

### Outstanding questions

* Are there any potential security implications here, especially on shared machines? Because any process could conceivably connect to any `govim` instance in the scenario presented and query information from Vim, file contents etc, which might subvert file permissions. Does `gopls` not have the same problem?

### Related `gopls` issues

* [x/tools/gopls: support advanced query syntax for Symbol](https://github.com/golang/go/issues/37237)
* [x/tools/gopls: fully qualify symbol names for Symbol](https://github.com/golang/go/issues/38273)
* [x/tools/gopls: consider a -utf8 flag to talk UTF8 offsets](https://github.com/golang/go/issues/38274)

cc @findleyr @leitzler

myitcv:
This enables govim to be used with fzf and other such Vim plugins that
require a client program in order to run a command. In the case of fzf,
the client instance of govim is used in order to re-query the parent
instance, which in turn calls on to gopls' workspace symbol search. This
allows the search results within fzf to be refined on each key stroke.

This approach, however, is not specific to fzf. Hence some configuration
of Vim/fzf is required. For a sample .vimrc configuration see:

https://github.com/govim/govim/wiki/vimrc-tips#-tip-use-fzf-with-workspace-symbol-search

Fixes #795

### DIFF
--- a/cmd/govim/cpprotocol.go
+++ b/cmd/govim/cpprotocol.go
@@ -1,0 +1,56 @@
+package main
+
+// TODO: perhaps all this child-parent stuff is better housed in an internal helper
+// package
+
+// Child-parent protocol
+//
+// The child-parent protocol (better name required) defines the process by
+// which a child and parent instances of govim communicate. The principal use
+// case is being able to write scripts that run against an editor instance. The
+// motivating example here being fzf which uses a child instances to query
+// gopls (via the parent) for session symbol information.
+//
+// The parent instance exposes the command that should be run in order to
+// create that child instance that then connects to the parent. This is exposed
+// via the GOVIMParentCommand() Vim function.
+//
+// TODO: expose help information for the various commands etc
+//
+// The child and parent communicate using a Unix domain socket. Messages on the
+// wire are encoded JSON. The child starts by sending all of its non-flag args
+// to the parent. The parent is then responds with pairs of data. The pair
+// comprises a JSON-encoded number followed by a JSON-encoded value. The
+// following list of modes is supported:
+//
+// 0X - X is an integer representing the exit code the client should use
+// 1X - X is a string value that should be output to os.Stdout
+// 2X - X is a string value that should be output to os.Stderr
+// 3x - X is an interface{} value whose JSON representation should be output to os.Stdout
+// 4x - X is an interface{} value whose JSON representation should be output to os.Stderr
+//
+// If the parent instance encounters an error writing to the child's
+// connection, it assumes the child has closed the connected and it (the
+// parent) continues without error.
+
+type encodeCode int
+
+const (
+	encodeCodeExitCode encodeCode = iota
+	encodeCodeRawStdout
+	encodeCodeRawStderr
+	encodeCodeJSONStdout
+	encodeCodeJSONStderr
+)
+
+// Command is the interface implemented by parent (sub) comamnds
+type Command interface {
+	// Run is the implementation of the command. A Command is responseible for
+	// encoding its response (through sub commands) to the child.  This is
+	// typically done by all the commands in the hierarchy embedding the root
+	// *parentReq.
+	Run()
+
+	// Errorf is a convenience method for panic-ing with a knownParentError
+	Errorf(format string, args ...interface{})
+}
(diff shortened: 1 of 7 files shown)


cmd/govim/cpprotocol.go
```diff
@@ -0,0 +1,56 @@
+package main
+
+// TODO: perhaps all this child-parent stuff is better housed in an internal helper
```
leitzler, line 3: agree

cmd/govim/flag.go
```diff
@@ -17,12 +18,22 @@ func usage() {
 	fmt.Fprintf(os.Stderr, `
 Usage of govim:
 
-	govim [-tail] /path/to/gopls
+	govim [-tail] [-parent /path/to/uds] gopls ...
+
+See each command's help for more details, e.g.:
+
+   govim gopls -help
```
leitzler, line 25: Leftovers from an idea that wasn't implemented?

leitzler (approved):
Quite nice feature! Just nits, otherwise LGTM

cmd/govim/flag.go
myitcv: Yes :)

cmd/govim/cpprotocol.go
myitcv: Ok, will move as part of this PR.

myitcv: Actually, with your permission I will defer this until a later PR if that's ok?

PR #852, opened 2020-04-09
cmd/govim: persist highlights for explicit GOVIMHighlightReferences call

myitcv:
When manually invoking GOVIMHighlightReferences, it makes sense to
persist the highlights for as long as we can. To be safe we assume that
making any change to any .go file will potentially invalidate the
highlighted references (we might relax this at a later stage). We also
add the GOVIMClearReferencesHighlights command to explicitly remove the
highlights.

As part of this same commit we:

* move the definition of currentReferences to vimstate where it belongs
  (it is only ever accessed on the vimstate "thread")
* Tidy up and simplify some logic regarding removal of highlights

### DIFF
--- a/cmd/govim/buffer_events.go
+++ b/cmd/govim/buffer_events.go
@@ -79,6 +79,10 @@ type bufChangedChange struct {
 // bufChanged(bufnr, start, end, added, changes)
 //
 func (v *vimstate) bufChanged(args ...json.RawMessage) (interface{}, error) {
+	// For now any change (in a .go file) causes an existing highlights to be removed
+	v.highlightingReferences = false
+	v.removeReferenceHighlight(nil)
+
 	bufnr := v.ParseInt(args[0])
 	b, ok := v.buffers[bufnr]
 	if !ok {
--- a/cmd/govim/config/config.go
+++ b/cmd/govim/config/config.go
@@ -275,8 +275,13 @@ const (
 	CommandSuggestedFixes Command = "SuggestedFixes"
 
 	// CommandHighlightReferences highlights references to the identifier under
-	// the cursor.
+	// the cursor. The highlights are removed by a change to any file or a call
+	// to CommandClearReferencesHighlights.
 	CommandHighlightReferences Command = "HighlightReferences"
+
+	// CommandClearReferencesHighlights clears any highlighint of references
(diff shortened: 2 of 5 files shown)


cmd/govim/buffer_events.go
```diff
@@ -79,6 +79,10 @@ type bufChangedChange struct {
 // bufChanged(bufnr, start, end, added, changes)
 //
 func (v *vimstate) bufChanged(args ...json.RawMessage) (interface{}, error) {
+	// For now any change (in a .go file) causes an existing highlights to be removed
```
leitzler, line 82: How about adding this thing as something that only applies when having "manual" highlights? I.e.:
```go
// Nicely phrased comment about that this only applies to "manual mode"
if v.highlightingReferences {
    v.highlightingReferences = false
    v.removeReferenceHighlight(nil)
}
```

I would really prefer to have them removed as-is when using "automatic" mode.

Vim handles highlighting via text properties during buffer change very good so it feels like a step  backwards to clear them more than necessary. I don't have a bunch of examples that would degrade, except for adding spaces in front of a highlighted identifier, but I have a feeling that there are other cases of `bufChange` triggers that will accidentally swipe away my precious highlights :)

myitcv: `v.highlightingReferences` is only used to indicate we are in "manual mode". So resetting that here is correct.

Despite the current logic removing references when we re-enter "busy" state, we definitely want to remove them on the first change regardless of whether we are in "manual mode" or not. So I think the logic as it stands is correct. We also want to ensure they are removed when we re-enter the busy state, and that call remains.

If we take a look at the first three lines of `(*vimstate).removeReferenceHighlights` we see:

```go
	if v.highlightingReferences || v.currentReferences == nil {
		return nil
	}
```

`v.currentReferences == nil` implies there is no work to be done. If we look at the bottom of that same method:

```
	v.currentReferences = nil
```

i.e. once we have removed highlights, we zero the state on the `govim` side.

> but I have a feeling that there are other cases of bufChange triggers that will accidentally swipe away my precious highlights :)

But if you are relying on highlights to be shown when you enter non-busy state, the making a change will cause them to disappear in any case because you will re-enter the "busy" state.

PR #853, opened 2020-04-11
cmd/govim: add ExperimentalWorkaroundCompleteoptLongest config option

myitcv:


cmd/govim/config/config.go
```diff
@@ -212,6 +212,13 @@ type Config struct {
 	//
 	// Default: nil
 	ExperimentalCursorTriggeredHoverPopupOptions *map[string]interface{} `json:",omitempty"`
+
+	// ExperimentalWorkaroundCompleteoptLongest provides a partial workaround
+	// for users wanting completeopt+=longest like behaviour. There is a Vim
+	// bug, described in github.com/vim/vim/issues/5891, that prevents
+	// completeopt+=longest from working properly, and as the comments in that
+	// issue describe why the workaround needs to be in govim.
```
leitzler, line 220: Do we want a little bit more context here (if this is the only place people look at for the description)?
With regards to how `completeopt` can/must be configured. Like should the user also have `+=longest` and something about it implicitly requires `completeopt+=menu` (or `menuone`)?

myitcv: Good point. I'll make this more clear. I will save writing the explanation here and just put it into an improved comment.

PR #857, opened 2020-04-13
cmd/govim: do not remove highlights on delta when auto-highlighting

myitcv:
In 3139f797 I incorrectly removed all reference highlights when a delta
is received. However, the logic prior to that change was that highlights
would only be moved (even in insert mode) if the cursor falls outside of
a highlighted range.

Therefore, we adopt exactly the logic that @leitzler proposed in #852 to
only remove highlights on delta if we are in manual highlighting mode.


myitcv:
@leitzler approved above. Merging

PR #866, opened 2020-04-30
cmd/govim: allow duplicate items in completion candidates

myitcv:
When it comes to unimported completions, it is entirely possible that
two packages might have the same name. Generally however, we rely
entirely on gopls to "dedupe" the completion candidate list. So make
this change unconditionally is safe.


myitcv:
Tested locally. Coming up with a test for this is hard, because `gopls` currently returns a random list of at most 5 unimported candidates.

PR #883, opened 2020-06-09
ci: fix Vim master build Go version

myitcv:


myitcv:
Merging as fix is minimal. Tested by checking that the vim master workflow runs properly when run on pull request. 

Raised #884 for the test flake.

PR #885, opened 2020-06-10
cmd/govim: add support for "fill struct" code action

leitzler:
Adds a new command, "GOVIMFillStruct", that populate fields with zero
value in a struct using gopls new code action "Fill struct".

Currently it must be called on an empty struct with the cursor above the
right bracket '}', see golang/go#39506.

### DIFF
--- a/cmd/govim/fillstruct.go
+++ b/cmd/govim/fillstruct.go
@@ -1,0 +1,64 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"math"
+
+	"github.com/govim/govim"
+	"github.com/govim/govim/cmd/govim/internal/golang_org_x_tools/lsp/protocol"
+	"github.com/govim/govim/cmd/govim/internal/golang_org_x_tools/span"
+)
+
+func (v *vimstate) fillStruct(flags govim.CommandFlags, args ...string) error {
+	b, point, err := v.bufCursorPos()
+	if err != nil {
+		return fmt.Errorf("failed to determine cursor position: %v", err)
+	}
+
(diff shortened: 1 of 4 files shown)


cmd/govim/fillstruct.go
```diff
@@ -0,0 +1,54 @@
...
+			buri := b.URI()
+			if euri != buri {
+				return fmt.Errorf("got edits for file %v, but buffer is %v", euri, buri)
+			}
+			if ev := int(math.Round(dc.TextDocument.Version)); ev > 0 && ev != b.Version {
+				return fmt.Errorf("got edits for version %v, but current buffer version is %v", ev, b.Version)
+			}
+			edits := dc.Edits
+			if len(edits) != 0 {
+				if err := v.applyProtocolTextEdits(b, edits); err != nil {
```
myitcv, line 46: How many calls to `applyProtocolTextEdits` are you expecting here? I'm assuming the answer is one.

It might be slightly nicer to assert that and drop the `for` loops. Because otherwise it gives the impression at first reading that we're going to be calling `applyProtocolTextEdits` multiple times... which would be crazy because we'd want to gather all the edits first and then call once.

leitzler: Indeed, good idea, I updated it according to your suggestion!

It originated from the fact that we do get an array back from gopls. And since the feature is new and going to get a overview soon I went for the safe thing. If we (for some reason) get more than one edit we will end up throwing an error and rely on users reporting it as an issue (since our test probably won't cover an edge case).

PR #898, opened 2020-06-24
cmd/govim: use gopls serve command explicitly

myitcv:
In the near future, the default zero-command-implied serve behaviour
will disappear.


myitcv:
@ianthehat does this look good to you?

leitzler:
We do run into some issues with our `GOVIM_GOPLS_FLAGS` env var that I, for example, use to set `-remote=auto` (and `-ocagent`). Since these two flags aren't a part of `serve`, gopls fails to start.

@ianthehat how close is "the near future"?

I suggest that we revert this one until we have a nice solution for passing flags to `gopls`, unless the near future is too close :)

PR #901, opened 2020-07-03
deps: update x/tools and gopls to 0cc1aa72b347

akupila:
* gopls/internal/hooks, internal/lsp/source: only match full words in link regexes 0cc1aa72
* internal/lsp: support opening single files f01a4bec
* internal/lsp/regtest: add regression tests for on-disk file changes ea7be8d7
* gopls/doc/troubleshooting.md: update the gopls output channel name 7cb253f4
* internal/lsp/source: speed up completion candidate formatting 1837592e
* internal/lsp/cmd: change pre-filled issue header for gopls bug 9a0e0698
* internal/lsp: decorate error message from workspace/configuration a32c0cb1
* go/packages: move all overlay tests into overlay_test.go c138986d
* internal/lsp: remove VS Code-specific completion hack b2d8b033
* internal/lsp/cache: handle a few possible panics in PackageStats 4bdfe1a3
* internal/analysisinternal: prevent fillstruct panic on nil package aa3d5013
* internal/lsp/source: add a new symbolStyle configuration option aa94e735
* internal/lsp: fully qualify workspace Symbol matches 84cfedeb
* internal/lsp: add a new regtest to reproduce golang/go#39646 bcbc01e0
* internal/lsp/fake: explicitly set GOPACKAGESDRIVER=off in regtests 6506e20d
* internal/lsp: watch all files in the module and replace target fadf93ff


akupila:
_This is my first commit here, please let me know if I missed something._

This updates gopls to the latest version (`v0.4.2`). For me the primary goal was to fix gofmt, which was not working at all (panic in fillstruct on every save - fixed in https://github.com/golang/tools/commit/aa3d5013).

All changes in this commit were automatically generated:

```
go get golang.org/x/tools/gopls@latest
./_scripts/revendorToolsInternal.sh
```

Changelog was generated with:
```
git log 7480c7b4547d..0cc1aa72 --format="* %s %h"
```

leitzler:
Thanks a lot for your contribution!

It looks like `go.mod` had the deps updated to use [the release branch](https://github.com/golang/tools/compare/gopls-release-branch.0.4) (probably a result of running `go get ...@latest`), while the changelog and PR title points to `master` (i.e. commit `0cc1aa72b347`).

Would you mind updating so that it uses master instead? (i.e. `go get golang.org/x/tools@master golang.org/x/tools/gopls@master`).

The changelog looks good! Please add it directly to the commit message also, since that is what is added to the git history.

akupila:
@leitzler Thanks, that explains why the go.mod change was to the tag and not to the commit. I did what you suggested and it looks more in line with previous tools updates. The latest commit in tools is the same so the change log is still up to date.

PR #904, opened 2020-07-11
cmd/govim: add user enabled analyses configuration

leitzler:
As a part of gopls configuration the user can enable/disable specific
analyses.

A list of analyzers that can be enabled/disabled can be found in
[the gopls documentation](https://cs.opensource.google/go/tools/+/master:gopls/doc/analyzers.md) (note that the list isn't up to date):

Example:

`call govim#config#Set("Analyses", {"fillstruct": 1, "unreachable": 0})`

### DIFF
--- a/cmd/govim/config/config.go
+++ b/cmd/govim/config/config.go
@@ -151,6 +151,18 @@ type Config struct {
 	// GOFLAGS=-modfile=go.local.mod in order to use an alternative go.mod file.
 	GoplsEnv *map[string]string `json:",omitempty"`
 
+	// Analyses is a map of booleans (0 or 1 in VimScript) used to enable/disable
+	// specific analyses in gopls. Entries in the map are used to override
+	// defaults specified by gopls. A list of analyzers with their default value
+	// can be found in the gopls documentation (e.g.
+	// https://cs.opensource.google/go/tools/+/master:gopls/doc/analyzers.md for
+	// master).
+	//
+	// Example: govim#config#Set("Analyses", {"fillstruct": 1, "unreachable": 0})
+	//
+	// Default: nil
+	Analyses *map[string]bool `json:",omitempty"`
+
 	// ExperimentalAutoreadLoadedBuffers is used to reload buffers that are
 	// changed outside vim even when they are loaded (e.g. running two vim
 	// sessions in the same workspace). This is achieved by running "checktime"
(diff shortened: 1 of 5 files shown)


cmd/govim/config/config.go
```diff
@@ -151,6 +151,14 @@ type Config struct {
 	// GOFLAGS=-modfile=go.local.mod in order to use an alternative go.mod file.
 	GoplsEnv *map[string]string `json:",omitempty"`
 
+	// Analyses is a map of booleans (0 or 1 in VimScript) used to enable/disable
+	// specific analyses in gopls.
```
myitcv, line 155: I think we need to expand this comment a bit to indicate that it's only the presence of an entry in the map that enables/disables an analysis. i.e. if I supplied an empty map for the `Analyses` config field then the default setting for each respective analysis would be used (according to the `gopls` docs).

Indeed it might even be worthwhile linking to the specific version of the `gopls` docs we are pinned to (although this could follow in a later PR)

cmd/govim/internal/vimconfig/vimconfig.go
```diff
@@ -87,6 +89,25 @@ func copyStringValMap(i, j *map[string]string) *map[string]string {
 	return &res
 }
 
+func copyBoolValMap(i *map[string]int, j *map[string]bool) *map[string]bool {
```
myitcv, line 92: I think this logic needs to be a merge, first applying `j`, then `i`, no? Otherwise if I supply a non-nil `i` then non of the defaults in `j` are applied.

leitzler: That was my initial though as well, then I noticed that the others (`copyStringValMap` and `copyMap`) didn't. Or I might have misunderstood `ToConfig(...)`, hm.

`j` are defaults , so in case of conflicting keys the ones in `i` should still override right?

cmd/govim/config/config.go
leitzler: Ack, I'll add a link to master for this PR.

cmd/govim/internal/vimconfig/vimconfig.go
myitcv: > then I noticed that the others (copyStringValMap and copyMap) didn't. Or I might have misunderstood ToConfig(...), hm.

I think that's just a case of neither the experimental popup options nor the env maps currently having defaults. If they did then we would need a merge. So more a case of me being lazy/writing for the situation at the time than anything. We could equally tidy them up in the interests of completeness though, you're right.

> j are defaults , so in case of conflicting keys the ones in i should still override right?

Exactly.

leitzler:
PTAL and see if this looks better :)

PR #907, opened 2020-07-13
ci: fix Vim master build

myitcv:
The Go version numbers used by the setup-go action do not,
unfortunately, following the versioning scheme used by Go itself. Fix
the version used for the Vim master build.


myitcv:
For comfort, this was tested in https://github.com/govim/govim/actions/runs/167151873

internal/cmd/genconfig/genconfig.go
```diff
@@ -228,6 +229,53 @@ func writeFileFromTmpl(path string, tmpl string, v interface{}) {
...
+)
+
+// semverGitHubGoVersion is a temporary hack to deal with the fact that the GitHub
+// action setup-go requires a semantic version, despite the fact that Go does
+// not follow semver.
+//
+// More discussion: https://github.com/actions/setup-go/issues/63
+//
+// This function panics in case we don't see a valid Go version, e.g.
+// go1.14beta1rc1
```
leitzler, line 244: It doesn't, but it do panic if `rc` comes before `beta`, e.g. `go1.14rc1beta1` :)

Do we really need the extra error handling? It feels a bit less readable to me, and since it "only" generate new code it is visible before commit if the output is totally off (and I doubt that using "go1.14beta1rc1" as input is likely?).

If we should have error handling at all, I think that more likely input would be things like `go1.15.0-beta1` which returns a more subtle error `1.15.0--beta1`, or `go1.15-.0-beta1` with returns the more obvious `1.15-.0-beta1`.

But on the other hand, it's a temporary hack and it gets the job done.

myitcv: Ha! Sloppy indeed. Thanks for the catch. We probably don't need the extra error handling as you say. I'll keep it simpler for now.

myitcv: Indeed I made the error handling slightly more robust... but it's still incomplete and hacky. Gets the job done for now. I floated the idea of a more robust alternative here https://github.com/golang/go/issues/32450#issuecomment-657653381

PR #913, opened 2020-07-26
deps: upgrade x/tools and gopls to 77f530d86f9a

leitzler:
* internal/lsp: handle deletion of file content 77f530d8
    * internal/lsp: move fillstruct suggested fixes out of analysis 4025ed84
    * internal/lsp: support extract function 6d307edf
    * internal/lsp: split only on the first = for Go environment variables 9cbb971a
    * internal/lsp: support go1.12 cf97b7f4
    * internal/lsp: don't re-run `go env` for every call to WriteEnv e6549b8d

    The update also updates the fillstruct logic according to recent changes
    in gopls where it now return Commands instead of direct edits.

    As gopls went from edits to commands, it will not be possible to
    fill more than one struct per fillstruct call.

    This also adds support for incoming LSP ApplyEdit calls, to be able
    to use LSP ExecuteCommand.


cmd/govim/fillstruct.go
```diff
@@ -34,33 +35,112 @@ func (v *vimstate) fillStruct(flags govim.CommandFlags, args ...string) error {
...
+		// version. Let's go for the first one and let the user call fillstruct again if they
+		// want to fill several structs on the same line.
+		ca := codeActions[0]
+		_, err := v.server.ExecuteCommand(context.Background(), &protocol.ExecuteCommandParams{
+			Command:                ca.Command.Command,
+			Arguments:              ca.Command.Arguments,
+			WorkDoneProgressParams: protocol.WorkDoneProgressParams{},
+		})
+		if err != nil {
+			v.Logf("executeCommand failed: %v", err)
```
myitcv, line 76: We should be relaying this back to the user who called `GOVIMFillStruct`

cmd/govim/fillstruct.go
```diff
@@ -34,33 +35,112 @@ func (v *vimstate) fillStruct(flags govim.CommandFlags, args ...string) error {
...
+		v.govimplugin.applyEditsCh = nil
+		v.govimplugin.applyEditsLock.Unlock()
+		close(done)
+		return nil
+	})
+
+	for {
+		select {
+		case <-done:
+			return nil
```
myitcv, line 89: Per above, return any error returned from `ExecuteCommand`

cmd/govim/fillstruct.go
```diff
@@ -34,33 +35,112 @@ func (v *vimstate) fillStruct(flags govim.CommandFlags, args ...string) error {
...
+		// verify that the version of the edits matches a buffer
+		var buf *types.Buffer
+		for _, b := range v.buffers {
+			if b.URI() != span.URI(dc.TextDocument.URI) {
+				continue
 			}
+
 			if ev := int(math.Round(dc.TextDocument.Version)); ev > 0 && ev != b.Version {
-				return fmt.Errorf("got edits for version %v, but current buffer version is %v", ev, b.Version)
+				v.Logf("got edits for buffer version %v, found matching buffer with version %v", ev, b.Version)
```
myitcv, line 123: This feels like it should be an error to me. i.e. we've managed to get ourselves into a state where we were given an edit for the wrong version of a document. There might be conditions under which "best efforts" is acceptable, but I'd want those to be very clearly identified and commented.

cmd/govim/fillstruct.go
```diff
@@ -34,33 +35,112 @@ func (v *vimstate) fillStruct(flags govim.CommandFlags, args ...string) error {
...
+
+			buf = b
 		}
-	}
 
-	if len(edits) != 0 {
-		if err := v.applyProtocolTextEdits(b, edits); err != nil {
-			return err
+		if buf == nil {
+			res.FailureReason = fmt.Sprintf("got edits for buffer %v, but didn't find it", dc.TextDocument.URI)
```
myitcv, line 131: Isn't it possible we will be asked to edit a file we don't have open?

Admittedly not for fill struct, so perhaps leave this as a TODO for later?

cmd/govim/gopls_client.go
```diff
@@ -151,9 +151,36 @@ func (g *govimplugin) Configuration(ctxt context.Context, params *protocol.Param
...
 	defer absorbShutdownErr()
-	panic("ApplyEdit not implemented yet")
+	g.logGoplsClientf("ApplyEdit: %v", pretty.Sprint(params))
+
+	var err error
+	var res *protocol.ApplyWorkspaceEditResponse
+	g.applyEditsLock.Lock()
+	if g.applyEditsCh == nil {
+		// ApplyEdit wasn't send by another blocking call so it's fine to just schedule the edits here
+		done := make(chan struct{})
```
myitcv, line 163: Where `g.applyEditsCh == nil`, then we don't need to block here. Simply `g.applyEditsLock.Unlock()` then schedule (there is no reason to hold the lock whilst scheduling). We definitely do not need to block until the edit itself is applied.

cmd/govim/gopls_client.go
```diff
@@ -151,9 +151,36 @@ func (g *govimplugin) Configuration(ctxt context.Context, params *protocol.Param
...
+	} else {
+		// There is an ongoing call that can apply edits on the vim thread. A Schedule here would deadlock so
+		// pass the edits to the vim thread instead.
+		e := applyEditCall{params: params, responseCh: make(chan applyEditResponse)}
+		g.applyEditsCh <- e
+		aer := <-e.responseCh
+		res = aer.res
+		err = aer.err
+	}
+	g.applyEditsLock.Unlock()
```
myitcv, line 180: Given the comment above about the `g.applyEditsCh == nil` branch this might well need to move.

cmd/govim/fillstruct.go
```diff
@@ -34,33 +35,112 @@ func (v *vimstate) fillStruct(flags govim.CommandFlags, args ...string) error {
...
+	v.govimplugin.applyEditsLock.Lock()
+	v.govimplugin.applyEditsCh = editsCh
+	v.govimplugin.applyEditsLock.Unlock()
+	done := make(chan struct{})
+
+	v.tomb.Go(func() error {
+		// We can only apply one command at the moment since they all target the same document
+		// version. Let's go for the first one and let the user call fillstruct again if they
+		// want to fill several structs on the same line.
+		ca := codeActions[0]
```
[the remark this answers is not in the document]
leitzler: I opted this mainly because multiple responses are an edge case, and this code block is a temporary thing until https://github.com/golang/go/issues/40438 is fixed. Since we pass `Only: [..] protocol.RefactorRewrite` we will only get `fillstruct` responses (afaik that's the only supported rewrite action for .go files).

What would the purpose of selecting by title be? If it is to ensure that the response is a `fillstruct` command we could instead look at the `Command` field within the command in the response. Example response:
```
Command:     &protocol.Command{
    Title:     "Fill foo with default values",
    Command:   "fill_struct",
    Arguments: {
        "{\"uri\":\"file:///private/var/folders/xm/ls208xd174v95pgd20rn_qbh0000gn/T/tmp.lmXRD9mP/main.go\",\"range\":{\"start\":{\"line\":7,\"character\":6},\"end\":{\"line\":7,\"character\":6}}}",
    },
},
```

If it is to ensure that we pick the "correct one" that has a matching name, we must then first extract the struct identifier in `govim`. I avoided that complexity in this PR since it would only be needed when there are multiple empty structs to fill on the same line.

cmd/govim/fillstruct.go
leitzler: Good point, was going for "best efforts" here but an error is better.

cmd/govim/fillstruct.go
leitzler: Had too much focus on just `fillstruct` and didn't think about that, I'll add a TODO.

cmd/govim/gopls_client.go
leitzler: Ah, right. I was thinking about the case where a user invoked `GOVIMFillStruct` the moment before we schedule edits in this branch. And missed the fact that it won't deadlock since it would be an entirely unrelated call.

PR #915, opened 2020-07-30
deps: upgrade x/tools and gopls to 6467de6f

myitcv:
* internal/lsp: remove Mod handles 6467de6f
* internal/lsp: handle nil pointer in go.mod links b47602a8


myitcv:
No changes to `govim` code here; merging.

PR #919, opened 2020-07-31
cmd/govim: change test skip to be based on golang.org/issues/36880

myitcv:


myitcv:
Turns out we can't remove the skip just yet; we got a flake. However it looks like the skip should now be a function of https://github.com/golang/go/issues/36880 based on https://github.com/golang/go/issues/37164#issuecomment-588623351. Will update the URL and merge.

PR #920, opened 2020-07-31
cmd/govim: disable flaky quickfix_empty_files and modfile_changes tests

myitcv:
Disable tests that are flaky on CI as a result of intermittent gopls
bugs, namely golang.org/issues/39646 and golang.org/issues/40269.

We have open issues against these bugs, so it doesn't make sense to
suffer the daily disappointment of a failing build against gopls@master,
or indeed have a sometimes-red (and therefore misleading) main build.


myitcv:
Merging on the basis the change is fairly uncontroversial (and will make the `main` build go green)

PR #922, opened 2020-08-01
cmd/govim: refactor popup menu selection handling

leitzler:
We only have one menu styled popup where the user can select one of the
entries, in suggested fixes. When adding support for other LSP features
as CodeLens we will need other types of menus.

This change moves suggested fixes related code into suggested_fixes.go
and prepare for other menus by making the popup selection code more
generic.

### DIFF
--- a/cmd/govim/main.go
+++ b/cmd/govim/main.go
@@ -197,6 +197,14 @@ type govimplugin struct {
 	applyEditsCh   chan applyEditCall
 	applyEditsLock sync.Mutex
 
+	// popupMenus is a map of active popups where the user can make a selection. It is keyed
+	// by popupID and the value is a function that will be called when the popup is closed.
+	// It is either closed by selecting an item (1-indexed selection is passed as argument),
+	// by calling popup_close() (0 is passed as argument) or ESC is pressed (-1 is passed as
+	// argument. Adding and removing popups is protected by popupMenusLock.
+	popupMenus     map[int]func(selection int) error
+	popupMenusLock sync.Mutex
+
 	bufferUpdates chan *bufferUpdate
 
 	// inShutdown is closed when govim is told to Shutdown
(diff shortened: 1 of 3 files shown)


cmd/govim/main.go
```diff
@@ -197,6 +197,14 @@ type govimplugin struct {
 	applyEditsCh   chan applyEditCall
 	applyEditsLock sync.Mutex
 
+	// popupMenus is a map of active popups where the user can make a selection. It is keyed
```
myitcv, line 200: Can you clarify, how can there be more than one active popup?

leitzler: Active is probably not the best word. What we need to handle is multiple ShowMessageRequest calls from gopls. And it could, in theory, happen during a "suggested fixes" selection. I still have to figure out the best way to queue them (or if we just should close any open menu popup if we receive a ShowMessageRequest call).

cmd/govim/vimstate.go
```diff
@@ -157,26 +157,16 @@ func (v *vimstate) popupSelection(args ...json.RawMessage) (interface{}, error)
 	v.Parse(args[0], &popupID)
 	v.Parse(args[1], &selection)
 
-	var edits []protocol.WorkspaceEdit
-	var ok bool
-	if edits, ok = v.suggestedFixesPopups[popupID]; !ok {
+	v.popupMenusLock.Lock()
+	cb, ok := v.popupMenus[popupID]
+	if !ok {
+		v.popupMenusLock.Unlock()
```
myitcv, line 163: Could just defer the unlock, no?

cmd/govim/main.go
```diff
@@ -197,6 +197,14 @@ type govimplugin struct {
...
+	// by popupID and the value is a function that will be called when the popup is closed.
+	// It is either closed by selecting an item (1-indexed selection is passed as argument),
+	// by calling popup_close() (0 is passed as argument) or ESC is pressed (-1 is passed as
+	// argument. Adding and removing popups is protected by popupMenusLock.
+	popupMenus     map[int]func(selection int) error
+	popupMenusLock sync.Mutex
```
myitcv, line 206: Can you clarify why a lock is required here? Because it appears that all work with the map happens on the vimstate "thread"

myitcv (commented):
Couple of questions

leitzler:
Won't be needed any more since the current way CodeLens works doesn't play well as a popup (no way to differentiate each item that might have same title etc.). Closing.

PR #932, opened 2020-08-26
cmd/govim: add support for popups showing WorkDoneProgress

leitzler:
Add experimental support for popups showing WorkDoneProgress.
The feature is disabled by default, and can be enabled via .vimrc:

`call govim#config#Set("ExperimentalProgressPopups", 1)`

When enabled govim will create notification style popups in
the upper right corner when gopls report progress of a task
(e.g. initial workspace load, running tests or "go generate").

The popup will close itself afterwards.

Note that this is an experimental feature. It has rather limited
use initially since govim doesn't yet support running tests via
gopls for example.

### DIFF
--- a/cmd/govim/gopls.go
+++ b/cmd/govim/gopls.go
@@ -137,6 +137,8 @@ func (g *govimplugin) startGopls() error {
 	initParams.Capabilities.Workspace.DidChangeConfiguration.DynamicRegistration = true
 	initParams.Capabilities.Workspace.DidChangeWatchedFiles.DynamicRegistration = true
 
+	initParams.Capabilities.Window.WorkDoneProgress = true
+
 	if _, err := g.server.Initialize(context.Background(), initParams); err != nil {
 		return fmt.Errorf("failed to initialise gopls: %v", err)
 	}
--- a/cmd/govim/gopls_client.go
+++ b/cmd/govim/gopls_client.go
@@ -12,6 +12,7 @@ import (
 	"github.com/govim/govim/cmd/govim/internal/golang_org_x_tools/lsp/protocol"
 	"github.com/govim/govim/cmd/govim/internal/golang_org_x_tools/lsp/source"
 	"github.com/govim/govim/cmd/govim/internal/golang_org_x_tools/span"
+	"github.com/govim/govim/cmd/govim/internal/types"
 	"github.com/kr/pretty"
 )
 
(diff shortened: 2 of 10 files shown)


cmd/govim/gopls.go
```diff
@@ -137,6 +137,8 @@ func (g *govimplugin) startGopls() error {
 	initParams.Capabilities.Workspace.DidChangeConfiguration.DynamicRegistration = true
 	initParams.Capabilities.Workspace.DidChangeWatchedFiles.DynamicRegistration = true
 
+	initParams.Capabilities.Window.WorkDoneProgress = true
```
myitcv, line 140: Is it possible to tell the server about new capabilities at runtime?

Ideally we would only say we can do `WorkDoneProgress` if we actually require it.

cmd/govim/gopls_client.go
```diff
@@ -225,11 +226,57 @@ func (g *govimplugin) PublishDiagnostics(ctxt context.Context, params *protocol.
...
+
+	var title string
+	if title, ok = raw["title"].(string); !ok && kind == "begin" { // required for "begin"
+		return fmt.Errorf("expected required field 'title' not set")
+	}
+
+	g.Schedule(func(govim.Govim) error {
+		v := g.vimstate
+
+		if v.config.ExperimentalProgressPopups == nil || !*v.config.ExperimentalProgressPopups {
```
myitcv, line 251: Per comment above, we don't actually need this check if we only enable progress as required.

Otherwise, if we can't drive this based on the capability, then I suggest we check the config value before scheduling (using the config lock - see elsewhere). It's relatively expensive to schedule only to do nothing.

cmd/govim/gopls_client.go
```diff
@@ -225,11 +226,57 @@ func (g *govimplugin) PublishDiagnostics(ctxt context.Context, params *protocol.
...
+	return nil
 }
+
 func (g *govimplugin) WorkDoneProgressCreate(ctxt context.Context, params *protocol.WorkDoneProgressCreateParams) error {
 	defer absorbShutdownErr()
-	panic("WorkDoneProgressCreate not implemented yet")
+	g.logGoplsClientf("WorkDoneProgressCreate callback: %v", pretty.Sprint(params))
+	g.Schedule(func(govim.Govim) error {
+		v := g.vimstate
+		if v.config.ExperimentalProgressPopups == nil || !*v.config.ExperimentalProgressPopups {
```
myitcv, line 270: Ditto

myitcv (approved):
LGTM with a couple of comments/questions.

cmd/govim/gopls.go
leitzler: `gopls` doesn't support that :(. The overhead of having progress callbacks instead of logmessage callbacks should be negligible.

cmd/govim/gopls_client.go
leitzler: Done, now checking config value before schedule.

cmd/govim/gopls_client.go
leitzler: Done

PR #939, opened 2020-09-09
cmd/govim/config: add SymbolMatcher and SymbolStyle config

myitcv:
These correspond to gopls configurations that can be set on the session.


leitzler (approved):
LGTM

Issue #931, opened 2020-08-25
cmd/govim: can't get the cursor back by pressing CTRL-O after GOVIMMotion


roy2220:
### What version of Vim/Gvim are you using?

<details><summary>Vim/Gvim version</summary><br><pre>
$ vim --version
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jul 28 2020 00:00:00)
Included patches: 1-1307
Modified by <bugzilla@redhat.com>
Compiled by <bugzilla@redhat.com>
Huge version without GUI.  Features included (+) or not (-):
+acl               -farsi             +mouse_sgr         +tag_binary
+arabic            +file_in_path      -mouse_sysmouse    -tag_old_static
+autocmd           +find_in_path      +mouse_urxvt       -tag_any_white
+autochdir         +float             +mouse_xterm       -tcl
-autoservername    +folding           +multi_byte        +termguicolors
-balloon_eval      -footer            +multi_lang        +terminal
+balloon_eval_term +fork()            -mzscheme          +terminfo
-browse            +gettext           +netbeans_intg     +termresponse
++builtin_terms    -hangul_input      +num64             +textobjects
+byte_offset       +iconv             +packages          +textprop
+channel           +insert_expand     +path_extra        +timers
+cindent           +ipv6              +perl/dyn          +title
-clientserver      +job               +persistent_undo   -toolbar
-clipboard         +jumplist          +popupwin          +user_commands
+cmdline_compl     +keymap            +postscript        +vartabs
+cmdline_hist      +lambda            +printer           +vertsplit
+cmdline_info      +langmap           +profile           +virtualedit
+comments          +libcall           -python            +visual
+conceal           +linebreak         +python3/dyn       +visualextra
+cryptv            +lispindent        +quickfix          +viminfo
+cscope            +listcmds          +reltime           +vreplace
+cursorbind        +localmap          +rightleft         +wildignore
+cursorshape       +lua/dyn           +ruby/dyn          +wildmenu
+dialog_con        +menu              +scrollbind        +windows
+diff              +mksession         +signs             +writebackup
+digraphs          +modify_fname      +smartindent       -X11
-dnd               +mouse             -sound             -xfontset
-ebcdic            -mouseshape        +spell             -xim
+emacs_tags        +mouse_dec         +startuptime       -xpm
+eval              +mouse_gpm         +statusline        -xsmp
+ex_extra          -mouse_jsbterm     -sun_workshop      -xterm_clipboard
[...]

PR #945, opened 2020-09-10
cmd/govim: remember previous position before GOVIMMotion call (#931)

roy2220:
See #931 for details

### DIFF
--- a/cmd/govim/motions.go
+++ b/cmd/govim/motions.go
@@ -131,6 +131,7 @@ func (v *vimstate) motion(args ...json.RawMessage) (interface{}, error) {
 	}
 
 	if targetNode != nil {
+		v.ChannelEx("normal! m'")
 		position := b.Fset.Position(resolv(targetNode))
 		v.ChannelCall("cursor", position.Line, position.Column)
 	}


myitcv:
Thanks for raising this, @roy2220 - I think we need to update the tests in https://github.com/govim/govim/blob/main/cmd/govim/testdata/scenario_default/motion.txt to reflect this change. Would you like to give that a shot?

roy2220:
> Thanks for raising this, @roy2220 - I think we need to update the tests in https://github.com/govim/govim/blob/main/cmd/govim/testdata/scenario_default/motion.txt to reflect this change. Would you like to give that a shot?

I am sorry to say that I have moved to [vim-lsc](github.com/natebosch/vim-lsc) recently, so I'm not going to make more efforts on this.

Govim is good, but vim-lsc seems to have a bright future ahead, for it's ambitious aim to build a nice generic LSP client.

It would be exhilarating if your are going to start up a new project such as `golsc`, LSP client witten in go, I would then love to contribute to it.

myitcv:
No problem, @roy2220. 

> for it's ambitious aim to build a nice generic LSP client.

This is, incidentally, the longer term goal of `govim` too, and to do so writing the LSP client in Go:

https://github.com/govim/govim/wiki/FAQ#why-should-i-use-govim-over-vim-govim-lsp-why-is-govim-not-language-agnostic

But per that FAQ the shorter term goal was/is to iterate quickly on the Go language with `gopls`. 

Enjoy `vim-lsc` (or did you mean `vim-lsp`?) and we may see you again soon 😄

theclapp:
Just because @roy2220 won't be using govim any more and thus won't be updating any tests, that doesn't mean that #931, which this fixes, isn't still a bug, or that this PR shouldn't move forward (arguably, tests or no tests).

myitcv:
@theclapp - agreed. I've just raised https://github.com/govim/govim/pull/1022 as a follow-up (couldn't update this PR because the original branch has been removed)

@roy2220 - thank you for raising the original issue and proposing a fix in this PR. I have credited you in my change.

PR #954, opened 2020-09-17
deps: update x/tools and gopls to 63098cc4

myitcv:
* internal/imports: don't set Context.WorkingDir, which was renamed 63098cc4
* internal/lsp: fix concurrency issues in view initialization 655488c8
* internal/lsp: pass options by reference instead of by value c537a342
* internal/lsp/cache: fix release tag parsing c9a70fc2
* internal/memoize: show key type in panics f128e626
* internal/lsp/mod: handle nil pointers in code lenses 587cf233
* cover: fix sorting of profile segment boundaries 56d9a0cd
* gopls/internal/regtest: simplify expectation return values c8d9e05b


myitcv:
No govim changes, will merge on green

PR #962, opened 2020-10-02
cmd/govim: fix line-wise visual selection in GOVIMStingFn

leitzler:
The line-wise mode always used two lines regardless of how many lines
that was selected. It is now possible to use GOVIMStringFn on more
lines.


myitcv (approved):
Excellent. Thank you

PR #966, opened 2020-10-05
cmd/govim: call didChangeWatchedFiles for go.sum files

### LINKED ISSUE #961: cmd/govim: send didChangeWatchedFiles notifications for go.sum files
I recently broke (and will soon fix...) the `scenario_modreadonly/script-config_set_env_goflags_mod_readonly` test.
While debugging this, I noticed that `govim` wasn't sending `didChangeWatchedFiles` notifications for `go.sum` files, which is required with `-mod=readonly` (otherwise, the Go command may fail with "updates to go.sum needed").

myitcv:
This is not easy to test for because we don't get any diagnostics etc
from gopls for go.sum files. So this has been manually "testing" for
now.

Fixes #961


leitzler (approved):
LGTM

PR #968, opened 2020-10-08
cmd/govim: support multi-line progress messages in popup

leitzler:
Progress messages do sometimes contain newlines. To ensure that the
popup is rendered correct the progress message is now divided into
multiple lines.

### DIFF
--- a/cmd/govim/progress.go
+++ b/cmd/govim/progress.go
@@ -2,17 +2,19 @@ package main
 
 import (
 	"sort"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


myitcv (approved):
LGTM with one question: can we add a test for this? i.e. is there a way we can trigger a progress message that we know will be multiline? I guess testing for the existence of that popup is hard...

leitzler:
It is tricky with the UI indeed. There aren't any multiline progress messages at the moment, as this is a gardening change for "go test" for example. It has been tested manually at least.

PR #971, opened 2020-10-09
cmd/govim: add support for ranged StringFn w/o visual mode

leitzler:
Vim supports ranged commands in different ways, and some of them doesn't
use visual mode at all. For example, ":%GOVIMStringFn
encoding/hex.EncodeToString" that encodes the entire file.

This change adds support for non-visual ranged GOVIMStringFn calls.


myitcv (approved):
LGTM - might be worth including a test with a specific range, prior to %?

leitzler:
The test case could use some love so I created a separate issue for it.

Issue #441, opened 2019-08-09
Support running 'go build' and 'go test' through govim


cespare:
In migrating from vim-go, one thing I miss are the :GoBuild / :GoTest commands. I had these bound to shortcuts and used them a lot.

I'm still experimenting to see if GOVIMQuickfixDiagnostics suffices instead of using :GoBuild (looks like maybe?), but either way I'd still like to have :GoTest.

Basically, the flow would be:

- Find the package the current file is in
- Run `go test` from that package dir
- Report failures in the quickfix window

This makes it easy to run the tests and jump directly to failures within vim.

An equivalent of vim-go's :GoTestFunc would be a great bonus feature. This figures out the test function that the cursor is in and then runs `go test -run '^TestSomeFunction$'`. This is handy if you have a lot of failures or if the tests are slow and you want to quickly run a single test.

tjcain:
I have been playing around with a `GOVIMTest` command this evening, is the below the kind of functionality required?

Tests pass:
![govim_test_pass](https://user-images.githubusercontent.com/37666423/77949961-c1904900-72bf-11ea-8ba0-47a11dcd549d.gif)

Tests fail:
![govim_test_fail](https://user-images.githubusercontent.com/37666423/77949982-c81ec080-72bf-11ea-9ef6-9dca303f4743.gif)

If so I am happy to clean up my hacking around and take this issue on, could also implement [#441](https://github.com/govim/govim/issues/441) in the same PR.

cespare:
Nice! Looks good.

> [...] could also implement [#441](https://github.com/govim/govim/issues/441) in the same PR.

That's this PR. Did you mean to refer to another one?

tjcain:
Ah, yes - sorry I meant to link in #442 !

myitcv:
@tjcain very happy for you to work on this.

One thing I'd encourage you to look at is the [code lens](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-15.md#textDocument_codeLens) feature of LSP. `gopls` has some support for this, so I'd want to be leverage that as far as possible.

tjcain:
Will do 👍

PR #974, opened 2020-10-15
cmd/govim: add support for testing via gopls GOVIMGoTest

leitzler:
Add command GOVIMGoTest that, via gopls, execute test and benchmark
functions.

It will execute the test func under the cursor, and also support
executing multiple test functions via visual mode or ranged command.

E.g.

Select range via visual mode and call `":'<,'>GOVIMGoTest"` to test all
functions that intersect with the selection.

`":%GOVIMGoTest"` tests the entire file
`":20,$GOVIMGoTest"` tests all functions from line 20 to the end

Also fixes a offset error when there are multiple progress popups open
at the same time.

Updates #441

### DIFF
--- a/cmd/govim/code_action.go
+++ b/cmd/govim/code_action.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+
+	"github.com/govim/govim"
+	"github.com/govim/govim/cmd/govim/internal/golang_org_x_tools/lsp/protocol"
+	"github.com/govim/govim/cmd/govim/internal/types"
+)
+
+func (v *vimstate) runGoTest(flags govim.CommandFlags, args ...string) error {
+	if c := v.config.ExperimentalProgressPopups; c == nil || !*c {
+		opts := make(map[string]interface{})
+		opts["mousemoved"] = "any"
+		opts["moved"] = "any"
+		opts["padding"] = []int{0, 1, 0, 1}
+		opts["wrap"] = true
+		opts["border"] = []int{}
+		opts["highlight"] = "ErrorMsg"
+		opts["line"] = 1
+		opts["close"] = "click"
+		v.ChannelCall("popup_create", []string{"GOVIMGoTest requires progress popups. Add this to your .vimrc:",
+			" call govim#config#Set(\"ExperimentalProgressPopups\", 1)"}, opts)
+		return nil
+	}
+	b, _, err := v.bufCursorPos()
+	if err != nil {
+		return fmt.Errorf("failed to get cursor position: %v", err)
+	}
+	start, end, err := v.rangeFromFlags(b, flags)
+	if err != nil {
+		return err
+	}
+
+	ca, err := v.server.CodeAction(context.Background(), &protocol.CodeActionParams{
+		TextDocument: b.ToTextDocumentIdentifier(),
+		Range: protocol.Range{
+			Start: start.ToPosition(),
+			End:   end.ToPosition(),
+		},
+		Context: protocol.CodeActionContext{
+			Only: []protocol.CodeActionKind{protocol.GoTest},
+		},
+		WorkDoneProgressParams: protocol.WorkDoneProgressParams{},
+	})
+	if err != nil || len(ca) == 0 {
+		return err
+	}
+	if len(ca) > 1 {
+		return fmt.Errorf("got %d CodeActions, expected no more than 1", len(ca))
+	}
+
+	c := ca[0]
+
+	token := protocol.ProgressToken(fmt.Sprintf("govim%d", rand.Uint64()))
+	if _, ok := v.progressPopups[token]; ok {
+		return fmt.Errorf("failed to init progress, duplicate token")
+	}
+	v.progressPopups[token] = &types.ProgressPopup{}
+
+	_, err = v.server.ExecuteCommand(context.Background(), &protocol.ExecuteCommandParams{
+		Command:   c.Command.Command,
+		Arguments: c.Command.Arguments,
+		WorkDoneProgressParams: protocol.WorkDoneProgressParams{
+			WorkDoneToken: token,
+		},
+	})
+
+	return err
+}
--- a/cmd/govim/config/config.go
+++ b/cmd/govim/config/config.go
@@ -341,6 +341,19 @@ const (
 	//
 	// Note that this feature is experimental in gopls.
 	CommandGCDetails Command = "GCDetails"
+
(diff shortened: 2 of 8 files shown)


leitzler:
[asciicast](https://asciinema.org/a/ppo5FsfHFUyYvd0twLkk9WeVk)

cespare:
Thanks for working on this!

I agree with @myitcv -- I think the output should be in a quickfix window, not a popup. For one thing, you ought to be able to easily jump from the error output to the line where the error occurred, which is a key part of the quickfix functionality. @myitcv makes some other good points, too; a popup is too ephemeral for this use case (I might want to close the window and reopen it, say) and I suspect the popup will become unwieldy if there is a lot of test output.

I have a question about running individual tests: if I want to run 2 tests, is it possible to run a single `go test` invocation with both of them, rather than two separate `go test`s? Is that some limitation of gopls?

leitzler:
Thanks for your input!
The popup is a part of the `Progress` support in `gopls` (or rather LSP) and a generic feature where I think that a popup isn't a bad idea per se. It could possibly just be a progress bar instead, but most things in `gopls` doesn't report any percent so it feels tricky to show something good UX-wise. That part is however opt-in by setting `ExperimentalProgressPopup` in `vimrc`.

With that said, if you do not opt in to that feature there isn't a way to see if a test pass today, it is completely silent. So that is where input and ideas are extra valuable!

> I agree with @myitcv -- I think the output should be in a quickfix window, not a popup.

This is a good idea. One thing about this is that output from `gopls` is just plain text and not structured in any way (`gopls` runs `-v -count=1 -run ^funcname$`). It feels a bit dirty to parse the output (even though I know that test2json is doing that). Should we only add entries to the quickfix list if the test failed? What about the rest of the output?

> I have a question about running individual tests: if I want to run 2 tests, is it possible to run a single go test invocation with both of them, rather than two separate go tests? Is that some limitation of gopls?

Yeah, that is how `gopls` [works at the moment](https://github.com/golang/tools/blob/a78bdc4e06a3ebf144292e01d9400b396495c0b6/internal/lsp/command.go#L290-L292), one invocation per func.

cespare:
> Should we only add entries to the quickfix list if the test failed? What about the rest of the output?

Personally, I don't want to see any output if the tests pass, other than some indication that it succeeded.

(I think this is how vim-go does it.)

leitzler:
> Personally, I don't want to see any output if the tests pass, other than some indication that it succeeded.

Any ideas how that can be indicated? (best option I have right now is a `:echom`).

cespare:
> best option I have right now is a `:echom`

Yeah that seems reasonable and I think that's what vim-go does.

myitcv:
> Any ideas how that can be indicated?

FWIW, I think a popup, top right (or similar) would be well suited to this. Particular if it was coloured in some way.

e.g. test finishes, status popup changes to green. Disappears after 3 seconds (e.g. like a Bootstrap alert)

leitzler:
> FWIW, I think a popup, top right (or similar) would be well suited to this. Particular if it was coloured in some way.

Without the actual test output? Something like this?

[asciicast](https://asciinema.org/a/bT0b6RedB8zdJtHnEMWWa4sYc)

(The border looks super weird in asciinema only).

As a bit of extra context, the actual text `gopls` sends is verbatim:
- `Running...` at the start
- `gopls_test: completed` at the end

And then the test output as it appears in between as separate progress messages.

leitzler:
Another thing that might be good to know is that the big red popup that shows for a **failed** test is something we will get regardless.

The reason is that it is sent as a stand-alone and unrelated asynchronous `ShowMessage` with severity `Error`.

myitcv:
> Without the actual test output?

I think so (by default at least). Unless I'm running a test with `-v` then I don't care for the output. It's only when a test fails that I care.

So the top right update popup could say:

```
Running test XYZ...
```

and then update to green on pass with:

```
Test XYZ passed
```

or fail and show the error output in a split.

> Another thing that might be good to know is that the big red popup that shows for a failed test is something we will get regardless.

That does not feel right at all. We have invoked a command here, are controlling the display of progress. I don't think `gopls` should then "take over" and decide how we view the test failure.

leitzler:
Yeah, I would like to have the name of the test there as well, but it isn't known to `govim` unless we parse the test output. `govim` only tells `gopls` a file & range to test and isn't aware of any names.

> That does not feel right at all. We have invoked a command here, are controlling the display of progress. I don't think gopls should then "take over" and decide how we view the test failure.

Totally agree!

I think that the rationale behind this is that the `progress` part of LSP doesn't carry any severity. I.e. `gopls` cannot tell us if the test passed or failed using `progress`.

The way it does is via `ShowMessage` with severity `Error` when failed and `Info` when it pass (and `govim` filters `Info` so it isn't visible).

leitzler:
Progress reporting ought to be generic to work well with other things, like when `go generate` is added. How about this completely different idea:

- Keep the progress popup, but limit the height to 10(?) lines
- Create a new hidden scratch buffer (just in-memory)
- Push all of the progress output to that buffer

It could look something like this:
[asciicast](https://asciinema.org/a/BLVFdRwzrYQKqCT7DmSj9v4Hp)

myitcv:
I like this latest suggestion. I think the popup size top right should be fixed (because otherwise the "dance" as new content arrives will be distracting). 

The approach of having it "scroll past" also works well.

Not least because in the case of failure we could have a command that opens that buffer in a split for the user to review, in full. 

I'm still not convinced about `gopls` doing a `ShowMessage` on test failure, but perhaps we should actually try it in practice in order to a) provide more concrete feedback and b) hopefully stir up different ideas on how it could work.

leitzler:
This is how it looks like currently, please share any thoughts!

I ran it on a "real" test file, from the Grafana project, just to see how it behaves in a more life like situation.

The huge red popup showing isn't super, and we can't suppress it since it is a completely unrelated ShowMessage callback from gopls.

When calling `GOVIMLastProgress` (naming suggestions are very welcome) vim creates and jumps to a new scratch buffer named "gopls-progress-<timestamp>".

[asciicast](https://asciinema.org/a/9y320P0Nv2qyr0bvn246iAsWi)

myitcv:
This looks like a great start! Fantastic that we can prototype these things so easily.

> The huge red popup showing isn't super, and we can't suppress it since it is a completely unrelated ShowMessage callback from gopls.

Agreed. I also saw your discussion with @findleyr on this on Slack. Sounds like discussion is moving forward on that front?

myitcv (approved):
LGTM - please squash

PR #976, opened 2020-10-17
cmd/govim: fix overlap and formating issues in progress popups

leitzler:
Two small fixes for progress I caught while experimenting with "go
test":

- Multiple progress popups did overlap
- The progress message could intentionally start with space, we
  shouldn't trim those


myitcv (approved):
LGTM

PR #980, opened 2020-10-23
cmd/govim: improve qf index selection on update

tbruyelle:
GoVim often looses the selected qf entry, which results in setting qf index to 1.

This is especially annoying when you have a lot of warnings (from an other package opened in another buffer), and you try to fix compile errors. As soon as you fix the error, the qf list index becomes 1 and you have to scroll down the warnings to find again your compile errors.

Actually, the selected qf entry is kept only if you're not fixing it, or if your fix doesn't affect its line number.

I tried to patch this behavior a little bit, by selecting the closest entry with the same filename (or dirname), if the selected entry isn't found. Thus, the next qf index stays in the same file or package, so you can continue to fix compile errors. I tested a little bit and I found the behavior a way more convenient than before.

An other improvement would be to fallback on entries with a compile errors rather than warnings or less.

I'm not sure this is the kind of change expected for qf index selection, because I didn't understand the comment located above my change : 

```go
        // In the future we might want to improve this logic
	// by stashing the last selected diagnostic when we flip to, for example,
	// references mode. But for now we keep it simple.
```

Also I didn't find any tests related to qf index selection, and I don't know how I can verify the qf index from the tests. Any help would be appreciated.

### DIFF
--- a/cmd/govim/quickfix.go
+++ b/cmd/govim/quickfix.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"path"
 	"path/filepath"
 
 	"github.com/govim/govim"
@@ -75,19 +76,45 @@ func (v *vimstate) updateQuickfixWithDiagnostics(force bool, wasPrevNotDiagnosti
 	if !wasPrevNotDiagnostics && len(v.lastQuickFixDiagnostics) > 0 {
 		var want qflistWant
 		v.Parse(v.ChannelExpr(`getqflist({"idx":0})`), &want)
-		if want.Idx == 0 {
+		if want.Idx == 1 {
+			// Don't guess selection if position is 1
 			goto NewIndexSet
 		}
 		wantIdx := want.Idx - 1
 		if len(v.lastQuickFixDiagnostics) <= wantIdx {
 			goto NewIndexSet
 		}
-		currFix := v.lastQuickFixDiagnostics[want.Idx-1]
+		currFix := v.lastQuickFixDiagnostics[wantIdx]
+		var fileNextIdx, fileLastIdx, dirFirstIdx int
 		for i, f := range fixes {
+			if f.Filename == currFix.Filename {
+				// Track index of the last entry of currFix file
+				fileLastIdx = i + 1
+				if fileNextIdx == 0 && f.Lnum >= currFix.Lnum {
+					// Track index of next entry of currFix file
+					fileNextIdx = i + 1
+				}
+			}
+			if dirFirstIdx == 0 && path.Dir(f.Filename) == path.Dir(currFix.Filename) {
+				// Track index of the first entry of currFix directory
+				dirFirstIdx = i + 1
+			}
 			if currFix.equalModuloBuffer(f) {
 				newIdx = i + 1
 				break
 			}
+		}
+		if newIdx == 0 {
+			// If currFix isn't found, set index to the next entry from the same file
+			newIdx = fileNextIdx
+		}
+		if newIdx == 0 {
+			// If fileNextIdx isn't set, set index to the last entry from the same file
+			newIdx = fileLastIdx
+		}
+		if newIdx == 0 {
+			// If fileLastIdx isn't set, set index to the first entry from the same directory
+			newIdx = dirFirstIdx
 		}
 	}
 NewIndexSet:
--- a/cmd/govim/quickfix.go
+++ b/cmd/govim/quickfix.go
@@ -76,8 +76,9 @@ func (v *vimstate) updateQuickfixWithDiagnostics(force bool, wasPrevNotDiagnosti
 	if !wasPrevNotDiagnostics && len(v.lastQuickFixDiagnostics) > 0 {
 		var want qflistWant
 		v.Parse(v.ChannelExpr(`getqflist({"idx":0})`), &want)
-		if want.Idx == 1 {
-			// Don't guess selection if position is 1
+		if want.Idx == 0 {
+			// NOTE: this should never happen since idx == 0 only if the qf is empty
+			// and we just tested it isn't.
 			goto NewIndexSet
 		}
 		wantIdx := want.Idx - 1
--- a/cmd/govim/quickfix.go
+++ b/cmd/govim/quickfix.go
@@ -82,9 +82,6 @@ func (v *vimstate) updateQuickfixWithDiagnostics(force bool, wasPrevNotDiagnosti
 			goto NewIndexSet
 		}
 		wantIdx := want.Idx - 1
-		if len(v.lastQuickFixDiagnostics) <= wantIdx {
-			goto NewIndexSet
-		}
 		currFix := v.lastQuickFixDiagnostics[wantIdx]
 		var fileNextIdx, fileLastIdx, dirFirstIdx int
 		for i, f := range fixes {
--- a/cmd/govim/testdata/scenario_default/quickfix_fallback_directory.txt
+++ b/cmd/govim/testdata/scenario_default/quickfix_fallback_directory.txt
@@ -5,14 +5,6 @@
 vim ex 'e charly.go'
 vimexprwait errors.golden.orig GOVIMTest_getqflist()
 
-# Verify we have the first entry selected
-# NOTE: disabled because when errors are spreaded accross multiple files like 
-# in this test, the quickfix list can be populated multiple times, which makes
-# GOVIM retains the previous selected entry.
-#vim expr 'getqflist({\"idx\": 0})'
-#stdout '{"idx":1}'
-#! stderr .+
-
 # Now move to error 'errc' and check the position
 vim expr 'setqflist([], \"r\", {\"idx\": 3})'
 vim expr 'getqflist({\"idx\": 0})'
--- a/cmd/govim/testdata/scenario_default/quickfix_fallback_directory.txt
+++ b/cmd/govim/testdata/scenario_default/quickfix_fallback_directory.txt
@@ -1,0 +1,140 @@
+# Test that the quickfix window logic fallbacks to first error of the selected
+# entry directory, if this one is not found in the new quickfix list and if
+# there's no other errors in this entry file.
+
+vim ex 'e charly.go'
+vimexprwait errors.golden.orig GOVIMTest_getqflist()
+
+# Verify we have the first entry selected
+vim expr 'getqflist({\"idx\": 0})'
+stdout '{"idx":1}'
+! stderr .+
+
+# Now move to error 'errc' and check the position
+vim expr 'setqflist([], \"r\", {\"idx\": 3})'
+vim expr 'getqflist({\"idx\": 0})'
+stdout '{"idx":3}'
+! stderr .+
+
+# Now fix selected error and assert the index points to the first file of the directory
+vim ex 'call cursor(3,1)'
+vim normal dd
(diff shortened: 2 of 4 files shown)


myitcv:
Thanks for raising this, @tbruyelle. This annoys me too!

> Actually, the selected qf entry is kept only if you're not fixing it, or if your fix doesn't affect its line number.

We do actually have a test, but, almost exactly as you observe, we are only checking that the selected index of the quickfix window is retained when we add a new quickfix entry (i.e. new error):

https://github.com/govim/govim/blob/4b9752946c43d4c3f4235af52ef2f2e74ffb29b9/cmd/govim/testdata/scenario_default/quickfix_retain_position.txt

So it's entirely possible (and indeed we have both observed) that when we remove an entry that we do not retain the position.

> Also I didn't find any tests related to qf index selection, and I don't know how I can verify the qf index from the tests. Any help would be appreciated.

💯   for asking about this! I think we should update the test linked above. For some background details on this style of test, please see [the test wiki](https://github.com/govim/govim/wiki/govim-tests). 

Just to refresh my intuition on the logic behind what should happen to the selected index in the quickfix list (in diagnostic mode):

* If the quickfix list changes in some way but the currently selected item is still there in the new list, we should keep it selected
* If the quickfix list changes in some way such that the currently selected item is _not_ there in the new list, then we fallback to finding the "closest next error"

"Closest next error" is defined as:

* next (defined in terms of file offset of the currently selected error) error in the same file, else
* last error in the same file, else
* first error in first file (lexicographic ordering) of files in the same directory

How does that sound to you?

cc @leitzler

tbruyelle:
Hi @myitcv,

Thanks for your response and your constructive feedback.

I'm completely fine with your definition of "Closest next error". I will try to implement it and complete the tests accordingly.

cmd/govim/quickfix.go
```diff
@@ -75,20 +76,46 @@ func (v *vimstate) updateQuickfixWithDiagnostics(force bool, wasPrevNotDiagnosti
 	if !wasPrevNotDiagnostics && len(v.lastQuickFixDiagnostics) > 0 {
 		var want qflistWant
 		v.Parse(v.ChannelExpr(`getqflist({"idx":0})`), &want)
-		if want.Idx == 0 {
+		if want.Idx == 1 {
```
tbruyelle, line 79: @myitcv I'm not sure if it was a bug to check 0 instead of 1, but I need that check to have consistant test results.

If we try to guess the new index when `want.Idx == 1`, we have flaky results because the qf list is populated multiple times, starting with sub-parts of the final list.

tbruyelle:
@myitcv Please find a new version of the PR, with your specifications.

I created 3 new tests for each cases when the index selection fallback on
- the next entry of the file
- the last entry of the file
- the first entry of the file directory

cmd/govim/quickfix.go
myitcv: The key is a comment a few lines above (which is easy to miss, I accept)

```
// Note: indexes are 1-based, hence 0 means "no index"
```

Hence, the check here is saying "if there is nothing selected, let's leave it that way"

cmd/govim/quickfix.go
```diff
@@ -75,20 +76,46 @@ func (v *vimstate) updateQuickfixWithDiagnostics(force bool, wasPrevNotDiagnosti
...
+			// Don't guess selection if position is 1
 			goto NewIndexSet
 		}
 		wantIdx := want.Idx - 1
 		if len(v.lastQuickFixDiagnostics) <= wantIdx {
```
myitcv, line 84: I really can't remember what condition this is catching. @tbruyelle do you have any intuition? If so, please can you add a comment? Otherwise, I think we should drop this conditional and associated block

myitcv (approved):
This looks good, @tbruyelle - thanks very much for working on this.

Just a couple of comments/questions to address.

The only overarching question I have is whether we could/should combine the three tests into one? That's certainly not critical though, and I'd be happy to leave it for a later PR.

cmd/govim/quickfix.go
tbruyelle: Indeed, the value is 0 when the quickfix is empty, but as soon as there's at least one entry, the command returns a number >0.

But since before you check `len(v.lastQuickFixDiagnostics) > 0`, I'm pretty sure the command will never return 0 here.

I'll try to explain why skipping selection when `want.Idx == 1` matters, using as an example the test `quickfix_fallback_directory.txt`. This tests uses 4 files, all of them has an error :

- `alice/alice.go`
- `bob.go`
- `charly.go`
- `dave.go`

For unknown reason (maybe a bug), the method `updateQuickfixDiagnostics` is invoked 2 times when a file is opened in the test (`vim ex 'e charly.go'`). The first call doesn't have the full list of errors (it's random, can be any error of the 4), while the second call has the full list of errors.

Example:
- first call : the qf list is populated with only 
```json
[{"filename": "bob.go"}]
```
- second call: the qf list is populated with the full list 
```json
[{"filename": "alice/alice.go"}, {"filename": "bob.go"}, {"filename": "charly.go"}, {"filename": "dave.go"}]
```
(I skipped the other fields for clarity)

If we don't skip index selection when currentIndex == 1 (`want.Idx == 1`, we probably need to rename the variable), the second call will try to select the index using `bob.go` as a reference, and so will select index 2. This is why without this condition, the test fails from the start at line 10, when we assert that opening vim with a file with errors the default qf index is 1.

If you want to experience it yourself, just replace `want.Idx==1` by `want.Idx==0`, and run the test `quickfix_fallback_directory` multiple times, it will randomly fail according to what is the content of the sublist of errors from the first call. It will succeed only if by any chance the subpart contains the same first item as the full list (`alice/alice.go`).

cmd/govim/quickfix.go
tbruyelle: At first look, I thought it was there to skip index selection when the current index is higher than the number of new fixes, but actually it doesn't compare with the new list length but with the previous list length...

I don't know if this situation can happen actually. I put a panic inside the `if` and run the tests, no errors, so I think we probably can remove it :)

tbruyelle:
> The only overarching question I have is whether we could/should combine the three tests into one? That's certainly not critical though, and I'd be happy to leave it for a later PR.

I can try to merge them 3 (maybe even 4 with `quickfix_retain`), it will result in a very big file with a lot of `errors1.golden.updated`, `errors2.golden.updated`, ...
If it's OK for you that's fine, maybe you have a specific pattern to name golden files in that case ?

cmd/govim/quickfix.go
myitcv: Let's drop it and see what breaks :) Easy to then add a fix back along with a comment!

myitcv:
> If it's OK for you that's fine, maybe you have a specific pattern to name golden files in that case ?

Let's leave it for now then. If we can clearly see at a later date that a single file is better, then yes, we can move to a single (big) file with appropriately named golden files (any pattern is fine). It's not clearly better to have a single file, so I'd rather stick with what you have for now to reduce churn on the change.

cmd/govim/quickfix.go
myitcv: > But since before you check len(v.lastQuickFixDiagnostics) > 0, I'm pretty sure the command will never return 0 here.

I'd tend to agree. Please can you add a comment to this effect? So that if we are wrong, we can fix the code _and_ the comment?

> I'll try to explain why skipping selection when want.Idx == 1 matters

Thanks. I think I followed your explanation; but please correct me if anything that follows misses the points you were making.

> For unknown reason (maybe a bug), the method updateQuickfixDiagnostics is invoked 2 times when a file is opened in the test (vim ex 'e charly.go'). The first call doesn't have the full list of errors (it's random, can be any error of the 4), while the second call has the full list of errors.

To my mind this could just as easily happen in "normal" operation, where I'm making changes that result in an error earlier in the first (or in a file lexicographically before the file position that is the first error). i.e. there isn't anything surprising/unique about the scenario to it being a "bug".

> If we don't skip index selection when currentIndex == 1 (want.Idx == 1, we probably need to rename the variable), the second call will try to select the index using bob.go as a reference, and so will select index 2.

I think what you're saying though assumes the user hasn't somehow changed the selection in the quickfix window. Because if they have, and they intentionally chose the diagnostic at index 1 (i.e. the first one), then _not_ retaining that selection when something new comes along would be "surprising". 

Again, this comes back to a point you made earlier:

> Indeed, the value is 0 when the quickfix is empty, but as soon as there's at least one entry, the command returns a number >0.

We can't distinguish between the user having intentionally chosen the first entry from what happens automatically when there is anything in the quickfix list. Hence I think we need to retain the currently selected item. It's at least the simplest thing to explain to the user.

What are your thoughts?

myitcv:
Thanks again, @tbruyelle - I think I replied to all your replies.

cmd/govim/quickfix.go
tbruyelle: Ok thanks for the clarification about what I suspected to be a bug. I agree with you, having the index at 1 can also be intentional from the user. 

I reverted the change but I  had also to remove the first assertion on the qf index at the beginning of the test to prevent failure. I added comments explaining what we talked about.
See b3330e7d197ce8022b76b97a90ea69b83dc045f7

cmd/govim/quickfix.go
tbruyelle: done f7c8ec1

tbruyelle:
I think I addressed the remaining points @myitcv, feel free to correct my english spelling in the code comments, I know it's not very good.

I'm happy to contribute to GoVim, I love the project philosophy (golang and gopls as first class citizen), which immediately makes me switch from vim-go :) Thanks for creating this awesome project, it runs very well!

cmd/govim/testdata/scenario_default/quickfix_fallback_directory.txt
```diff
@@ -0,0 +1,143 @@
+# Test that the quickfix window logic fallbacks to first error of the selected
+# entry directory, if this one is not found in the new quickfix list and if
+# there's no other errors in this entry file.
+
+vim ex 'e charly.go'
+vimexprwait errors.golden.orig GOVIMTest_getqflist()
+
+# Verify we have the first entry selected
+# NOTE: disabled because when errors are spreaded accross multiple files like
```
myitcv, line 9: I would delete the commented out test, because as you say, we can't assert anything due to the random order of diagnostics arriving. But then I would keep a comment explaining how such a situation can come about. 

FYI: s/spreaded/spread/ in your sentence.

myitcv:
This looks great, @tbruyelle. Only one last small request - thank you for your patience!

> feel free to correct my english spelling in the code comments, I know it's not very good.

Your English is excellent. I easily make more spelling/grammatical errors than you 😄 

> I'm happy to contribute to GoVim, I love the project philosophy (golang and gopls as first class citizen), which immediately makes me switch from vim-go :) Thanks for creating this awesome project, it runs very well!

Delighted to have you contribute. Thank you for taking the time and care over this change.

cmd/govim/testdata/scenario_default/quickfix_fallback_directory.txt
tbruyelle: ok removed e33cf50

tbruyelle:
@myitcv Thanks for your kind words and your attention.

myitcv:
Thanks again for working on this, @tbruyelle!

PR #983, opened 2020-10-28
cmd/govim: rename "vimstate.popupWinId" to please staticcheck

leitzler:
This change renames the field "popupWinId" in vimstate to "popupWinID"
to avoid getting diagnostics about it when running with staticcheck
enabled.


myitcv (approved):
LGTM

PR #986, opened 2020-11-06
ci: drop use of set-env and add-path

myitcv:
Per https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


leitzler (approved):
LGTM

PR #995, opened 2020-11-28
cmd/govim: add new config that changes how last progress is opened

leitzler:
When calling :GOVIMLastProgress govim creates a new vim buffer. This
change adds the possibility to change how the new buffer is opened.

It also changes the default from "e" to "below 10split".

Change by setting OpenLastProgressWith to a string value, e.g:
   "" - Do not open the new buffer
   "e" - Open in same window
   "split" - Open in split window
   "belowright 15vsplit" - Open in vertical split, right side, 15 columns

### DIFF
--- a/autoload/govim/config.vim
+++ b/autoload/govim/config.vim
@@ -148,6 +148,9 @@ function! s:validAnalyses(v)
   return [v:true, ""]
 endfunction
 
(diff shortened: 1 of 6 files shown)


myitcv (approved):
LGTM

PR #998, opened 2020-11-30
cmd/govim: fix gopls codelens setting

stamblerre:
https://golang.org/cl/273308 fixes a misspelling, which causes a govim test to fail.

### DIFF
--- a/cmd/govim/gopls_client.go
+++ b/cmd/govim/gopls_client.go
@@ -30,7 +30,7 @@ const (
 	goplsVerboseOutput        = "verboseOutput"
 	goplsEnv                  = "env"
 	goplsAnalyses             = "analyses"
-	goplsCodeLens             = "codelens"
+	goplsCodeLenses           = "codelenses"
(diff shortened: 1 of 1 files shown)


leitzler:
Thank you! We are a little bit in a "chicken and egg" situation here since this PR depends on `govim` updating tools to a version that includes CL273308, and updating is blocked by golang/go#42837.

I took a look at CL273308 and it looks like having the old name shouldn't break the tests to fail (yet), as per Heschis comment, which `govim` test is it that doesn't work?

stamblerre:
The failing test is TestScripts/scenario_default/suggested_fixes -- it doesn't expect the pop-up that says the setting is deprecated and replaced. Do you know if https://golang.org/cl/274241/ resolved https://github.com/golang/go/issues/42837?

leitzler:
I see! Don't know yet, GitHub Actions seems to have a major outage.. I'll update you both when I know!

myitcv:
This change appears to have been made elsewhere so closing.

PR #1009, opened 2020-12-19
deps: update x/tools and gopls to ae774e97

leitzler:
* internal/lsp: don't show duplicate diagnostics for go.mod errors ae774e97
* internal/lsp/source: only show "run file benchmarks" if available 5b06639e
* gopls/internal/regtest: test metadata validation only on save for go.mod 11a5667e
* go/analysis/passes/fieldalignment: filter comments from suggested fix 5b43ef93
* gopls/internal/regtest: test that accepting fix removes diagnostics 4b31ac34
* internal/lsp, gopls: recover from go-diff panics 008e4774
* internal/lsp: add titles to `go mod tidy` and update go.sum fixes 48e5bd11


leitzler:
No govim changes, merging.

PR #1011, opened 2020-12-31
Support "goimports-gofmt" value for FormatOnSave

bbrodriges:
As stated in godoc `FormatOnSave` accepts `goimports-gofmt` const. But this value has not been added to validation function thus after adding line to `vimrc`:

```
call govim#config#Set("FormatOnSave", "goimports-gofmt")
```
exception occures:

```
E605: Exception not caught: Tried to set invalid value for key FormatOnSave: must be one of: ['', 'gofmt', 'goimports']
```

This PR fixes above mentioned issue.

### DIFF
--- a/autoload/govim/config.vim
+++ b/autoload/govim/config.vim
@@ -35,7 +35,7 @@ function! s:pushConfig()
 endfunction
 
 function! s:validFormatOnSave(v)
-  let valid = ["", "gofmt", "goimports"]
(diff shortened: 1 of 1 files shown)


myitcv:
Thanks for the catch, @bbrodriges. This further adds weight to the argument that the entire config space should be defined in a single source of truth and code generated from there (including documentation and tests). But that is very much for a later day!

myitcv (approved):

Issue #1012, opened 2021-01-03
cmd/govim: ability to move popup window to top right


jbpratt:
### What version of Vim/Gvim are you using?

<details><summary>Vim/Gvim version</summary><br><pre>
❯ vim --version
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Nov 15 2020 16:46:37)
Included patches: 1-1989
Compiled by Arch Linux
Huge version without GUI.  Features included (+) or not (-):
+acl               -farsi             +mouse_sgr         +tag_binary
+arabic            +file_in_path      -mouse_sysmouse    -tag_old_static
+autocmd           +find_in_path      +mouse_urxvt       -tag_any_white
+autochdir         +float             +mouse_xterm       +tcl/dyn
-autoservername    +folding           +multi_byte        +termguicolors
-balloon_eval      -footer            +multi_lang        +terminal
+balloon_eval_term +fork()            -mzscheme          +terminfo
-browse            +gettext           +netbeans_intg     +termresponse
++builtin_terms    -hangul_input      +num64             +textobjects
+byte_offset       +iconv             +packages          +textprop
+channel           +insert_expand     +path_extra        +timers
+cindent           +ipv6              +perl/dyn          +title
-clientserver      +job               +persistent_undo   -toolbar
-clipboard         +jumplist          +popupwin          +user_commands
+cmdline_compl     +keymap            +postscript        +vartabs
+cmdline_hist      +lambda            +printer           +vertsplit
+cmdline_info      +langmap           +profile           +virtualedit
+comments          +libcall           +python/dyn        +visual
+conceal           +linebreak         +python3/dyn       +visualextra
+cryptv            +lispindent        +quickfix          +viminfo
+cscope            +listcmds          +reltime           +vreplace
+cursorbind        +localmap          +rightleft         +wildignore
+cursorshape       +lua/dyn           +ruby/dyn          +wildmenu
+dialog_con        +menu              +scrollbind        +windows
+diff              +mksession         +signs             +writebackup
+digraphs          +modify_fname      +smartindent       -X11
-dnd               +mouse             -sound             -xfontset
-ebcdic            -mouseshape        +spell             -xim
+emacs_tags        +mouse_dec         +startuptime       -xpm
+eval              +mouse_gpm         +statusline        -xsmp
+ex_extra          -mouse_jsbterm     -sun_workshop      -xterm_clipboard
+extra_search      +mouse_netterm     +syntax            -xterm_save
[...]

PR #1013, opened 2021-01-05
cmd/govim: fix panic when using custom hover opts "col" / "line"

leitzler:
Setting either ExperimentalCursorTriggeredHoverPopupOptions or
ExperimentalMouseTriggeredHoverPopupOptions fields "col" or "line"
did result in a panic.

The value was parsed as a raw json message, but since it is already
parsed once before it paniced. We now type assert the value instead.

Updates #1012

### DIFF
--- a/cmd/govim/hover.go
+++ b/cmd/govim/hover.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"strings"
 
 	"github.com/govim/govim/cmd/govim/config"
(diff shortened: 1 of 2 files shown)


cmd/govim/hover.go
```diff
@@ -113,16 +115,11 @@ func (v *vimstate) showHover(posExpr string, opts map[string]interface{}, userOp
 			opts[k] = v
 		}
 		var line, col int64
-		var err error
-		if lv, ok := opts["line"]; ok {
-			if line, err = rawToInt(lv); err != nil {
-				return nil, fmt.Errorf("failed to parse line option: %v", err)
-			}
+		if lv, ok := opts["line"].(float64); ok {
+			line = int64(lv)
```
myitcv, line 119: I think we probably want `math.Round` here? Otherwise it truncates towards zero.

Alternatively we could flip the decoding of the opts to use [`UseNumber`](https://pkg.go.dev/encoding/json#Decoder.UseNumber) (which might actually be a nice change, separate from this one)

myitcv (approved):
LGTM modulo one suggestion. 

Could also add a small test to cover this edge case?

cmd/govim/hover.go
leitzler: Interesting, would it ever be an issue to not use `math.Round` if the user always specify integers?

I wouldn't expect anyone to ever specify anything outside like, the range -10 to 10. Since it is just for line and column I would be surprised if anyone tried to set "line 1.7" :)

myitcv: The string value that represents the number gets converted to a `float`. There is no guarantee that the integer number represented by the string can be represented precisely as a float. Hence the truncation towards zero could cause an off by one if the values ends up being x.9999999 for example.

I'd prefer to use `UseNumber()` across the board for this reason, but for now `math.Round` is a sufficient workaround.

leitzler: I see! It wasn't an objection, was only out of curiosity :)

PR #1020, opened 2021-01-25
Support DirectoryFilters config option for gopls

bbrodriges:
`gopls` has an option to [`DirectoryFilters`](https://github.com/golang/tools/blob/e9a2c0706a5b9cc8ee5965d5a50c996ffb255a7b/internal/lsp/source/options.go#L213). It allows to add or remove certain directories from workspace. This option is especially useful if workspace is a root of giant monorepository project. Without this option it is impossible to open such projects due to endless indexing process.

This PR adds support for this option via standard `govim` config option.

### DIFF
--- a/cmd/govim/config/config.go
+++ b/cmd/govim/config/config.go
@@ -175,6 +175,18 @@ type Config struct {
 	// GOFLAGS=-modfile=go.local.mod in order to use an alternative go.mod file.
 	GoplsEnv *map[string]string `json:",omitempty"`
 
+	// GoplsDirectoryFilters can be used to exclude unwanted directories from the
+	// workspace. By default, all directories are included. Filters are an
+	// operator, `+` to include and `-` to exclude, followed by a path prefix
+	// relative to the workspace folder. They are evaluated in order, and
+	// the last filter that applies to a path controls whether it is included.
+	// The path prefix can be empty, so an initial `-` excludes everything.
+	//
+	// For more details and examples see:
+	//
+	// https://github.com/golang/tools/blob/master/gopls/doc/settings.md#directoryfilters-string
+	GoplsDirectoryFilters *[]string `json:",omitempty"`
+
 	// Analyses is a map of booleans (0 or 1 in VimScript) used to enable/disable
 	// specific analyses in gopls. Entries in the map are used to override
 	// defaults specified by gopls. A list of analyzers with their default value
--- a/autoload/govim/config.vim
+++ b/autoload/govim/config.vim
@@ -136,6 +136,18 @@ function! s:validGoplsEnv(v)
   return [v:true, ""]
 endfunction
 
+function! s:validGoplsDirectoryFilters(v)
+  if type(a:v) != 3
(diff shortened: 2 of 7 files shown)


bbrodriges:
Side note: I have added this feature by "copying" `GoplsEnv` option code. Please point if there are some places that I've missed.

I also have very little understanding of testdata scenarios mechanics. I'd be very grateful if you could help me to add proper tests for this feature.

cmd/govim/config/config.go
```diff
@@ -163,6 +163,19 @@ type Config struct {
...
+	// workspace. By default, all directories are included. Filters are an
+	// operator, `+` to include and `-` to exclude, followed by a path prefix
+	// relative to the workspace folder. They are evaluated in order, and
+	// the last filter that applies to a path controls whether it is included.
+	// The path prefix can be empty, so an initial `-` excludes everything.
+	//
+	// Examples:
+	// Exclude node_modules: `-node_modules`
+	// Include only project_a: `-` (exclude everything), `+project_a`
+	// Include only project_a, but not node_modules inside it: `-`, `+project_a`, `-project_a/node_modules`
```
myitcv, line 176: Please can you link here to official docs?

cmd/govim/config/config.go
```diff
@@ -163,6 +163,19 @@ type Config struct {
...
 
+	// GoplsDirectoryFilters can be used to exclude unwanted directories from the
+	// workspace. By default, all directories are included. Filters are an
+	// operator, `+` to include and `-` to exclude, followed by a path prefix
+	// relative to the workspace folder. They are evaluated in order, and
+	// the last filter that applies to a path controls whether it is included.
+	// The path prefix can be empty, so an initial `-` excludes everything.
+	//
+	// Examples:
+	// Exclude node_modules: `-node_modules`
```
myitcv, line 174: Not specific to your PR, rather the choice on the `gopls` side. Is it possible to ignore any `node_modules` directory, regardless of where it appears?

I would have thought the [`.gitignore` approach](https://git-scm.com/docs/gitignore) might have been better for this reason.

myitcv (approved):
Thanks for looking into this.

One question, related to the underlying support in `gopls` rather than this PR.

And one small change to be made.

cmd/govim/config/config.go
myitcv: Done in a follow up.

cmd/govim/config/config.go
myitcv: Will follow up on this question in a `gopls` issue.

myitcv:
@bbrodriges - thanks for kick starting this PR and doing the bulk of the work for this. I've added a quick test to verify things work as expected.

myitcv:
> I also have very little understanding of testdata scenarios mechanics. I'd be very grateful if you could help me to add proper tests for this feature.

I've added some more detail to https://github.com/govim/govim/wiki/govim-tests to explain the structure of our `testscript` tests.

PR #1026, opened 2021-01-31
cmd/govim: support quickfix code actions returning commands

### LINKED ISSUE #1025: cmd/govim: support commands in quickfix code action responses
Current main, 5e65049

When using the "suggested fixes" feature, `:GOVIMSuggestedFixes`, `govim` call gopls method `textDocument/codeAction` with CodeActionKind `QuickFix`. The response from `gopls` is shown to the user as a popup with one entry per suggested fix.

Currently `govim` only care about direct edits, the `edit` field in a code action response. But `gopls` might respond with a command instead, or even both.

According to LSP 3.16, the current version, we should support both of them in the same response:
> If a code action provides an edit and a command, first the edit is executed and then the command.

leitzler:
Code action responses can contain both WorkspaceEdit and Command
according to LSP 3.16, and edits should be applied before executing
the command.

We ignored the command field when handling suggested fixes. If a
suggested fix includes a command it is now executed after any edits are
applied.

Closes #1025

### DIFF
--- a/cmd/govim/vimstate.go
+++ b/cmd/govim/vimstate.go
@@ -62,7 +62,7 @@ type vimstate struct {
 	// suggestedFixesPopups is a set of suggested fixes keyed by popup ID. It represents
 	// currently defined popups (both hidden and visible) and have a lifespan of single
 	// codeAction call.
-	suggestedFixesPopups map[int][]protocol.WorkspaceEdit
+	suggestedFixesPopups map[int][]suggestedFix
 
 	// working directory (when govim was started)
 	// TODO: handle changes to current working directory during runtime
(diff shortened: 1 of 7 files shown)


cmd/govim/vimstate.go
```diff
@@ -196,9 +196,51 @@ func (v *vimstate) popupSelection(args ...json.RawMessage) (interface{}, error)
...
 
-	return nil, v.applyMultiBufTextedits(nil, edit.DocumentChanges)
+	// Edits should be applied before any Command according to LSP 3.16.
+	if len(fix.edit.DocumentChanges) > 0 {
+		if err := v.applyMultiBufTextedits(nil, fix.edit.DocumentChanges); err != nil {
+			return nil, err
+		}
+	}
+
+	if fix.command != nil {
```
myitcv, line 208: We should look to DRY up this pattern now that it's repeated here and in fillstruct. Or are you thinking we do that in a later PR?

myitcv (approved):
LGTM with one question/thought.

cmd/govim/vimstate.go
leitzler: I intentionally left it this way since I'm not sure how things will work when Rob's work with the "command API" is done.

Hopefully we can cut the entire code action turnaround in the fillstruct case. So, "yes, in a later PR" :)

PR #1027, opened 2021-02-01
cmd/govim: add ExperimentalAllowModfileModifications

myitcv:
ExperimentalAllowModfileModifications controls whether gopls should
automatically modify go.{mod,sum} as a side effect of its use of cmd/go
or not. Setting this option to true can be considered a gopls equivalent
to setting GOFLAGS=-mod=mod when using cmd/go. Leaving it unset or
setting it to false is equivalent to -mod=readonly. Indeed setting
GOFLAGS=-mod=XYZ is now considered deprecated as a means of controlling
this behaviour in gopls.

For reference: in Go 1.16 cmd/go build commands default to -mod=readonly
(equivalent to leaving this option unset, or setting it to false),
whereas prior to Go 1.16 the default was -mod=mod (equivalent to setting
this option to true).

golang.org/issue/44008 tracks confirmation that this gopls option will
not be removed until a better alternative is found.

### DIFF
--- a/autoload/govim/config.vim
+++ b/autoload/govim/config.vim
@@ -186,6 +186,10 @@ function! s:validExperimentalProgressPopups(v)
   return s:validBool(a:v)
 endfunction
 
+function! s:validExperimentalAllowModfileModifications(v)
+  return s:validBool(a:v)
+endfunction
+
 let s:validators = {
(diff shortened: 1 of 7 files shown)


leitzler (approved):
LGTM

PR #1032, opened 2021-02-04
[WIP] cmd/govim: convert unimported completion test to use non-std pkg

myitcv:


myitcv:
I'm not going to get to this so closing.

Issue #1028, opened 2021-02-02
Doing a :vim search over lots (thousands) of Go files takes a huge amount of time and cpu


theclapp:
### What version of Vim/Gvim are you using?

<details><summary>Vim/Gvim version</summary><br><pre>
$ vim --version
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Dec 27 2020 10:21:52)
macOS version
Included patches: 1-2164
Compiled by lmc@larrymbp14.local
Huge version with MacVim GUI.  Features included (+) or not (-):
+acl               +file_in_path      -mouse_sysmouse    -tag_old_static
+arabic            +find_in_path      +mouse_urxvt       -tag_any_white
+autocmd           +float             +mouse_xterm       -tcl
+autochdir         +folding           +multi_byte        +termguicolors
-autoservername    -footer            +multi_lang        +terminal
+balloon_eval      +fork()            -mzscheme          +terminfo
+balloon_eval_term +fullscreen        +netbeans_intg     +termresponse
+browse            +gettext           +num64             +textobjects
++builtin_terms    -hangul_input      +odbeditor         +textprop
+byte_offset       +iconv             +packages          +timers
+channel           +insert_expand     +path_extra        +title
+cindent           +ipv6              -perl              +toolbar
+clientserver      +job               +persistent_undo   +transparency
+clipboard         +jumplist          +popupwin          +user_commands
+cmdline_compl     +keymap            +postscript        +vartabs
+cmdline_hist      +lambda            +printer           +vertsplit
+cmdline_info      +langmap           +profile           +virtualedit
+comments          +libcall           -python            +visual
+conceal           +linebreak         -python3           +visualextra
+cryptv            +lispindent        +quickfix          +viminfo
-cscope            +listcmds          +reltime           +vreplace
+cursorbind        +localmap          +rightleft         +wildignore
+cursorshape       -lua               -ruby              +wildmenu
+dialog_con_gui    +menu              +scrollbind        +windows
+diff              +mksession         +signs             +writebackup
+digraphs          +modify_fname      +smartindent       -X11
+dnd               +mouse             -sound             -xfontset
-ebcdic            +mouseshape        +spell             +xim
+emacs_tags        +mouse_dec         +startuptime       -xpm
+eval              -mouse_gpm         +statusline        -xsmp
+ex_extra          -mouse_jsbterm     -sun_workshop      -xterm_clipboard
+extra_search      +mouse_netterm     +syntax            -xterm_save
[...]

PR #1034, opened 2021-02-06
cmd/govim: do not add short-lived buffers during :vimgrep

leitzler:
Using `:vimgrep` could result in a lot of short-lived buffers that is
sent to gopls since each file searched will trigger `BufReadPost`
autocommands.

Use `QuickFixCmdPre` & `QuickFixCmdPost` to detect when vimgrep runs,
and only add new buffers that still exists when vimgrep is done.
(see https://groups.google.com/g/vim_dev/c/qpu2O8cvprY/m/ZGAaOOlJBAAJ
for additional context)

Updates #1028

### DIFF
--- a/cmd/govim/buffer_events.go
+++ b/cmd/govim/buffer_events.go
@@ -18,7 +18,19 @@ import (
 
 func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
 	nb := v.currentBufferInfo(args[0])
+	if v.vimgrepPendingBufs != nil {
+		// We are getting BufRead autocommands during a vimgrep.
 
+		// Save the buffer without adding it yet since vimgrep could open
+		// a lot of files that are closed immediately, and we only want to
+		// add buffers still open when vimgrep is done.
+		v.vimgrepPendingBufs[nb.Num] = nb
+		return nil
+	}
+	return v.addBuffer(nb)
+}
+
+func (v *vimstate) addBuffer(nb *types.Buffer) error {
 	// If we load a buffer that already had diagnostics reported by gopls, the buffer number must be
 	// updated to ensure that sign placement etc. works.
 	diags := *v.diagnosticsCache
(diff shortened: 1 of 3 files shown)


cmd/govim/buffer_events.go
```diff
@@ -65,6 +77,29 @@ func (v *vimstate) bufReadPost(args ...json.RawMessage) error {
...
+}
+
+func (v *vimstate) bufQuickFixCmdPost(args ...json.RawMessage) error {
+	defer func() { v.vimgrepPendingBufs = nil }()
+
+	v.quickfixIsDiagnostics = v.quickfixIsDiagnostics && len(v.vimgrepPendingBufs) == 0
+
+	// Vim versions older than 8.2.2185 did not notify when buffers opened during vimgrep
+	// closed so we need to explicitly check if any of the buffers are still open.
+	for _, b := range v.vimgrepPendingBufs {
```
myitcv, line 92: Nit: is this worth doing in a batch? Not sure how expensive this could become.

cmd/govim/vimstate.go
```diff
@@ -89,6 +89,17 @@ type vimstate struct {
 	// lastProgressText points to the text in the most recently created progress
 	// popup.
 	lastProgressText *strings.Builder
+
+	// vimgrepPendingBufs contain buffers read during a vimgrep quickfix command,
+	// keyed by buffer number.
+	// The purpose is to avoid sending DidOpen/DidClose notifications to gopls
+	// for all files that doesn't match the pattern.
+	// Current vim (8.2.2465) do reuse the buffer number for all non-matching
```
myitcv, line 97: "Current" has a fairly short-lived meaning. Rephrase perhaps?

cmd/govim/vimstate.go
```diff
@@ -89,6 +89,17 @@ type vimstate struct {
 	// lastProgressText points to the text in the most recently created progress
 	// popup.
 	lastProgressText *strings.Builder
+
+	// vimgrepPendingBufs contain buffers read during a vimgrep quickfix command,
```
myitcv, line 93: Can we perhaps link here to a Vim issue/discussion that answers whether this behaviour is expected or not?

myitcv (approved):
LGTM with a few suggestions/questions.

cmd/govim/vimstate.go
```diff
@@ -89,6 +89,17 @@ type vimstate struct {
...
+	// keyed by buffer number.
+	// The purpose is to avoid sending DidOpen/DidClose notifications to gopls
+	// for all files that doesn't match the pattern.
```
myitcv, line 96: Small nit. Phrasing needs a bit of a tweak here I think. AIUI we're trying to avoid sending DidOpen/DidClose notifications to gopls for files which contained no pattern match, correct? Saying "files that didn't (sic) match the pattern" is a bit ambiguous because it could be the filepath that is matching a pattern or its contents.

cmd/govim/buffer_events.go
leitzler: I did consider that but I don't think that it is worth doing. We expect it to be rather few matches, and the additional overhead for a single channel call is probably tiny compared to all the things happening when a buffer is added to govim. It would require two loops as `v.addBuffer(...)` do a channel calls as well that can't be batched. I'll leave it as is for now.

cmd/govim/vimstate.go
leitzler: Hm, it was mostly to keep some kind of reference to a known version, i.e. "All versions, at least up until ... (might change in the future)". I dropped the version reference completely. It can still be tracked back if needed based on commit dates.

cmd/govim/vimstate.go
leitzler: Done

cmd/govim/vimstate.go
leitzler: Done

PR #1037, opened 2021-02-08
cmd/govim: remove internal buffer on BufWipeout

### LINKED ISSUE #1033: cmd/govim: delete internal buffers on BufWipeout autocommands
Current main, 9b702de

Currently we are listening for `BufDelete` and clean up our internal state when we get it, but the help file states that we won't get it for buffers that is deleted and not loaded:

> Before deleting a buffer from the buffer list.
> The BufUnload may be called first (if the
> buffer was loaded).
> Also used just before a buffer in the buffer
> list is renamed.

We should do the same thing when we get a `BufWipeout` autocommand to make sure we don't keep things in memory that aren't in vim anymore. From `BufWipeout`:

> Before completely deleting a buffer.  The
> BufUnload and BufDelete events may be called
> first (if the buffer was loaded and was in the
> buffer list).  Also used just before a buffer
> is renamed (also when it's not in the buffer
> list).

leitzler:
There are cases where we might get a BufWipeout autocmd without the
BufDelete (most notably when running :vimgrep). When that happen we kept
the internal buffer in govim even if we had no use of it anymore.

When using :bwipe we do get both BufDelete and BufWipeout, but as far as
we know BufDelete will always appear before BufWipeout. Trying to remove
an internal buffer due to BufWipeout will not cause an error.

Fixes #1033


myitcv (approved):
LGTM - does this code get triggered by existing tests?

(we really should add code coverage)

leitzler:
Yes, it does, in tests that use `:bwipe`.

Issue #492, opened 2019-09-05
cmd/govim: race condition in file watcher


leitzler:
### What version of Vim/Gvim are you using?

<details><summary>Vim/Gvim version</summary><br><pre>
$ vim --version
VIM - Vi IMproved 8.1 (2018 May 18, compiled Jul 22 2019 00:43:56)
macOS version
Included patches: 1-1729
Compiled by pontus@MacBook-Pro.localdomain
Huge version without GUI.  Features included (+) or not (-):
+acl               -farsi             -mouse_sysmouse    -tag_any_white
+arabic            +file_in_path      +mouse_urxvt       -tcl
+autocmd           +find_in_path      +mouse_xterm       +termguicolors
+autochdir         +float             +multi_byte        +terminal
-autoservername    +folding           +multi_lang        +terminfo
-balloon_eval      -footer            -mzscheme          +termresponse
+balloon_eval_term +fork()            +netbeans_intg     +textobjects
-browse            -gettext           +num64             +textprop
++builtin_terms    -hangul_input      +packages          +timers
+byte_offset       +iconv             +path_extra        +title
+channel           +insert_expand     -perl              -toolbar
+cindent           +job               +persistent_undo   +user_commands
-clientserver      +jumplist          +postscript        +vartabs
+clipboard         +keymap            +printer           +vertsplit
+cmdline_compl     +lambda            +profile           +virtualedit
+cmdline_hist      +langmap           -python            +visual
+cmdline_info      +libcall           -python3           +visualextra
+comments          +linebreak         +quickfix          +viminfo
+conceal           +lispindent        +reltime           +vreplace
+cryptv            +listcmds          +rightleft         +wildignore
-cscope            +localmap          -ruby              +wildmenu
+cursorbind        -lua               +scrollbind        +windows
+cursorshape       +menu              +signs             +writebackup
+dialog_con        +mksession         +smartindent       -X11
+diff              +modify_fname      -sound             -xfontset
+digraphs          +mouse             +spell             -xim
-dnd               -mouseshape        +startuptime       -xpm
-ebcdic            +mouse_dec         +statusline        -xsmp
+emacs_tags        -mouse_gpm         -sun_workshop      -xterm_clipboard
+eval              -mouse_jsbterm     +syntax            -xterm_save
+ex_extra          +mouse_netterm     +tag_binary        
+extra_search      +mouse_sgr         -tag_old_static    
   system vimrc file: "$VIM/vimrc"
[...]

PR #1038, opened 2021-02-09
cmd/govim: add directory filtering to darwin file watcher

leitzler:
The file watcher in darwin is recursive so adding and removing
directories to watch was no-ops. One issue with that was that the
filtering in cmd/govim/watcher.go that made sure files in directories
that starts with '.' or '_' wasn't watched didn't apply in darwin.

A consequence of that was that the test suite failed on darwin.

We now apply a simple filter to added paths and suppress events for
files outside those. Note that we do want to filter rather than
configuring the file watcher to be non-recursive to avoid #492.

### DIFF
--- a/cmd/govim/internal/fswatcher/os_darwin.go
+++ b/cmd/govim/internal/fswatcher/os_darwin.go
@@ -7,19 +7,29 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/fsnotify/fsevents"
 	"gopkg.in/tomb.v2"
 )
 
-const fRemoved = fsevents.ItemRemoved | fsevents.ItemRenamed
-const fChanged = fsevents.ItemModified | fsevents.ItemChangeOwner
-const fCreated = fsevents.ItemCreated
+const (
+	fRemoved = fsevents.ItemRemoved | fsevents.ItemRenamed
+	fChanged = fsevents.ItemModified | fsevents.ItemChangeOwner
+	fCreated = fsevents.ItemCreated
+	fIsDir   = fsevents.ItemIsDir
+)
 
 type fswatcher struct {
 	eventCh chan Event
 	es      *fsevents.EventStream
+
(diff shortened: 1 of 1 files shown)


cmd/govim/internal/fswatcher/os_darwin.go
```diff
@@ -7,19 +7,29 @@ import (
...
+	fChanged = fsevents.ItemModified | fsevents.ItemChangeOwner
+	fCreated = fsevents.ItemCreated
+	fIsDir   = fsevents.ItemIsDir
+)
 
 type fswatcher struct {
 	eventCh chan Event
 	es      *fsevents.EventStream
+
+	// Darwin do recursive watching so we need to filter files in directories that
```
myitcv, line 28: Nit: s/do/does/

cmd/govim/internal/fswatcher/os_darwin.go
```diff
@@ -7,19 +7,29 @@ import (
...
+	// wasn't explicitly added. Note that it is desirable to watch recursively to avoid
```
myitcv, line 29: Nit: s/wasn't/weren't/

cmd/govim/internal/fswatcher/os_darwin.go
```diff
@@ -7,19 +7,29 @@ import (
...
+	// a data race (#492).
+	watched     map[string]bool // keyed by full path to directory
```
myitcv, line 31: Per our offline chant, I'm not clear why this is necessary. On Darwin, given the watcher _is_ recursive (which is a good thing) we don't need to do the `filepath.Walk` that adds recursive watchers. 

Instead, we can simply check on the event callback whether the event is for a file we care about. i.e. ignore `.` and `_` prefixed directories, as well as submodules. 

It's not perfect of course and there are loads of potential FS races... but that's the nature of the beast 😄

myitcv (commented):
Added a couple of questions/thoughts on how we could do this differently.

PR #1056, opened 2021-03-08
cmd/govim: add posibility to set GoImports timeout as config

leitzler:
This change adds a new config, `GoImportsTimeout` expressed as a
time.Duration string, e.g.

```
call govim#config#Set("GoImportsTimeout", "1s")
```

Setting this will cancel the goimports call during buffer save after one
second.

Default: `0` (no timeout)

### DIFF
--- a/autoload/govim/config.vim
+++ b/autoload/govim/config.vim
@@ -156,6 +156,10 @@ function! s:validGofumpt(v)
   return s:validBool(a:v)
 endfunction
 
+function! s:validGoImportsTimeout(v)
+  return s:validString(a:v)
+endfunction
+
 function! s:validExperimentalAutoreadLoadedBuffers(v)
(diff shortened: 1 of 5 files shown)


myitcv:
Per my comments in the Slack thread, my push back here is the usual:

* it feels like this is very doable in the server
* hence we should do it in the server so every client doesn't have to do it

https://gophers.slack.com/archives/CL8JDP1GD/p1615267402000300?thread_ts=1615193841.100000&cid=CL8JDP1GD (for context)

mvdan:
I also think there should be sensible timeouts for *all* actions that could lock up the UI, such as saving a file. For example, saving go.mod files right now takes many seconds for me, for whatever reason I haven't debugged yet. There was also https://github.com/golang/go/issues/45092 just a few days ago.

PR #1057, opened 2021-03-12
cmd/govim: refactor file watching

### LINKED ISSUE #492: cmd/govim: race condition in file watcher

leitzler:
The previous file watchers was broken in several ways, where the two
major was:
* A race condition in the non-darwin watcher (#492)
* Darwin recursively watched the entire module directory, disregarding
  sub-modules, testdata directories and directories prefixed by _ or .
  This caused the test suite to always fail on macOS.

This change:
* Moves the non-recursive handling into os_other.go
* Adds an OS independent function that filter events specifically
  disallowed by Go (i.e. "testdata", sub-modules and directories
  prefixed by _ or .
* Adds more tests to raise file watching cover

Closes #492

### DIFF
--- a/cmd/govim/internal/fswatcher/os_darwin.go
+++ b/cmd/govim/internal/fswatcher/os_darwin.go
@@ -1,3 +1,4 @@
+//go:build darwin
 // +build darwin
 
 package fswatcher
@@ -13,24 +14,32 @@ import (
 	"gopkg.in/tomb.v2"
 )
 
-const fRemoved = fsevents.ItemRemoved | fsevents.ItemRenamed
-const fChanged = fsevents.ItemModified | fsevents.ItemChangeOwner
-const fCreated = fsevents.ItemCreated
+const (
+	fRemoved = fsevents.ItemRemoved | fsevents.ItemRenamed
+	fChanged = fsevents.ItemModified | fsevents.ItemChangeOwner
+	fCreated = fsevents.ItemCreated
+)
 
 type fswatcher struct {
 	eventCh chan Event
 	es      *fsevents.EventStream
 }
 
-func New(gomodpath string, tomb *tomb.Tomb) (*FSWatcher, error) {
-	dirpath := filepath.Dir(gomodpath)
-	dev, err := fsevents.DeviceForPath(dirpath)
+// New creates a file watcher that provide events recursively for all files and directories
+// that aren't filtered by the watch filter. The root argument must be an existing directory,
+// in our case the module root. FSWatcher will not send events for any path (file or directory)
+// where the filter returns "true".
+func New(root string, filter watchFilterFn, logf logFn, tomb *tomb.Tomb) (*FSWatcher, error) {
+	if fi, err := os.Stat(root); err != nil || !fi.IsDir() {
+		return nil, fmt.Errorf("provided root %q must be an existing directory", root)
+	}
+	dev, err := fsevents.DeviceForPath(root)
 	if err != nil {
-		return nil, fmt.Errorf("failed to retrieve device for path %v: %v", dirpath, err)
+		return nil, fmt.Errorf("failed to retrieve device for path %v: %v", root, err)
 	}
 
 	es := &fsevents.EventStream{
-		Paths:   []string{dirpath},
+		Paths:   []string{root},
 		Latency: 200 * time.Millisecond,
 		Device:  dev,
 		Flags:   fsevents.FileEvents | fsevents.WatchRoot,
@@ -40,7 +49,7 @@ func New(gomodpath string, tomb *tomb.Tomb) (*FSWatcher, error) {
 
 	// fsevents returns paths relative to device root so we need
 	// to figure out the actual mount point
-	mountPoint, err := filepath.Abs(dirpath)
+	mountPoint, err := filepath.Abs(root)
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -68,6 +77,10 @@ func New(gomodpath string, tomb *tomb.Tomb) (*FSWatcher, error) {
 				event := events[i]
 				path := filepath.Join(mountPoint, event.Path)
 
+				if filter(path) {
+					continue
+				}
+
 				// Darwin might include both "created" and "changed" in the same event
 				// so ordering matters below. The "created" case should be checked
 				// before "changed" to get a behavior that is more consistent with other
@@ -89,8 +102,6 @@ func New(gomodpath string, tomb *tomb.Tomb) (*FSWatcher, error) {
 	return &FSWatcher{&fswatcher{eventCh, es}}, nil
 }
 
-func (w *fswatcher) Add(path string) error    { return nil }
-func (w *fswatcher) Remove(path string) error { return nil }
 func (w *fswatcher) Close() error {
 	w.es.Stop()
 	return nil
--- a/cmd/govim/internal/fswatcher/os_others.go
+++ b/cmd/govim/internal/fswatcher/os_others.go
@@ -1,9 +1,12 @@
+//go:build !darwin
 // +build !darwin
 
 package fswatcher
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/fsnotify/fsnotify"
 	"gopkg.in/tomb.v2"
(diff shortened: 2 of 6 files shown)


cmd/govim/internal/fswatcher/os_darwin.go
```diff
@@ -68,6 +71,10 @@ func New(gomodpath string, tomb *tomb.Tomb) (*FSWatcher, error) {
 				event := events[i]
 				path := filepath.Join(mountPoint, event.Path)
 
+				if skipDir(path) {
```
myitcv, line 74: Is this guaranteed to be a directory path we are passing here?

cmd/govim/watcher.go
```diff
@@ -165,6 +87,39 @@ func (m *modWatcher) watch() {
...
 }
 
+func eventFilter(root string) func(string) bool {
+	return func(path string) bool {
+		if fi, err := os.Stat(path); err == nil && !fi.IsDir() {
+			path = filepath.Dir(path)
+		}
+
+		// TODO: cache ignored directories to reduce amount of os.Stat(...) calls
+		rel := strings.TrimPrefix(path, root)
```
myitcv, line 97: I would personally do another `strings.TrimPrefix` after this line with `os.PathSeparator` to get you the relative path. Saves the rather obscure logic below of starting from 1. Also, it's my understanding (and I would guess yours too) that the path we will be passed is clean, correct?

cmd/govim/watcher.go
```diff
@@ -165,6 +87,39 @@ func (m *modWatcher) watch() {
...
+		parts := strings.Split(rel, string(os.PathSeparator))
+		if len(parts) == 0 {
+			return false
+		}
+
+		curr := parts[0]
+		for i := 1; i < len(parts); i++ {
+			dir := parts[i]
+			switch dir[0] {
+			case '.', '_':
```
myitcv, line 107: Can we link to the relevant part of the Go source code/docs for the reference on this list of conditions?

cmd/govim/watcher.go
```diff
@@ -165,6 +87,39 @@ func (m *modWatcher) watch() {
 	}
 }
 
+func eventFilter(root string) func(string) bool {
+	return func(path string) bool {
+		if fi, err := os.Stat(path); err == nil && !fi.IsDir() {
```
myitcv, line 92: What happens for _files_ that start with `.` or `_`? Are they notified?

myitcv (commented):
I'll have a closer look later, but just a couple of initial questions.

cmd/govim/internal/fswatcher/os_darwin.go
leitzler: No, it could be directories or files. The function name came from the fact that it did check the directory part of any path provided, but that's just an implementation detail. Maybe just calling it `filter(...)` is more clear.
```go
if filter(path) {
```

cmd/govim/watcher.go
leitzler: Interesting, that obscure logic looks like a left-over from something else. I think that the original intent was to avoid checking for the root `go.mod` file but that won't happen anyway. I'll update.

cmd/govim/watcher.go
leitzler: Good idea, will do.

cmd/govim/watcher.go
leitzler: Yes. I missed the part about files. Need to update so that the check applies to both directories and files.

cmd/govim/watcher.go
leitzler: As note to self - https://golang.org/cmd/go/#hdr-Package_lists_and_patterns

cmd/govim/internal/fswatcher/os_darwin.go
myitcv: Agreed

leitzler:
Updated per your comments, PTAL when you have time.

cmd/govim/internal/fswatcher/os_others.go
```diff
@@ -13,49 +16,157 @@ type fswatcher struct {
...
+func New(root string, filter watchFilterFn, logf logFn, tomb *tomb.Tomb) (*FSWatcher, error) {
 	mw, err := fsnotify.NewWatcher()
 	if err != nil {
 		return nil, fmt.Errorf("failed to create new watcher: %v", err)
 	}
 
+	// activeWatchers is a map keyed by directory paths for directories that has been
+	// added to watch. The bool value is used to dedup multiple remove events for the same
+	// directory from fsnotify (inotify sends two events when removing a dir, "DELETE_SELF"
+	// and "DELETE,ISDIR" where the difference isn't exposed by fsnotify).
```
myitcv, line 31: Please can you explain in this comment _how_ the bool does that? i.e. what function does the use of a bool serve?

myitcv, line 31: Please can you explain in this comment _how_ the bool does that? i.e. what function does the use of a bool serve? That means you're not jumping below to find the reason :)

cmd/govim/internal/fswatcher/os_darwin.go
```diff
@@ -13,17 +14,19 @@ import (
...
+)
 
 type fswatcher struct {
 	eventCh chan Event
 	es      *fsevents.EventStream
 }
 
-func New(gomodpath string, tomb *tomb.Tomb) (*FSWatcher, error) {
-	dirpath := filepath.Dir(gomodpath)
+func New(root string, filter watchFilterFn, logf logFn, tomb *tomb.Tomb) (*FSWatcher, error) {
```
myitcv, line 28: Can `root` be a filepath?

Would be good to get some docs for both `New()` functions on what `root` and `filter` do, and the values that `root` can take.

cmd/govim/internal/fswatcher/os_darwin.go
```diff
@@ -13,17 +14,19 @@ import (
...
+	dirpath := filepath.Dir(root)
```
myitcv, line 29: We seem to calculate `filepath.Dir` here, but from what I can tell `root` will only ever be a directory? 

And why do we end up watching the parent directory of `root`? A comment here would be useful too.

cmd/govim/internal/fswatcher/os_others.go
```diff
@@ -13,49 +16,157 @@ type fswatcher struct {
...
+	w := &FSWatcher{&fswatcher{
+		eventCh: eventCh,
+		errCh:   mw.Errors,
+		mw:      mw,
+		logf:    logf,
+	}}
+
+	var populateWatchers func(string) ([]string, error)
+	populateWatchers = func(initPath string) (files []string, err error) {
+		return files, filepath.Walk(initPath, func(path string, info os.FileInfo, err error) error {
```
myitcv, line 44: I'm not such a fan of `files` being populated as a side effect of the second return value here; it hurts readability to my mind. 

Can `populateWatchers` be a method on `fswatcher` - would that tidy things up a bit?

cmd/govim/watcher.go
```diff
@@ -165,6 +87,59 @@ func (m *modWatcher) watch() {
...
+			if _, err := os.Stat(filepath.Join(root, curr, "go.mod")); err == nil {
+				return true
+			}
+		}
+		return false
+	}
+}
+
 func ofInterest(path string) bool {
 	return filepath.Ext(path) == ".go" || filepath.Base(path) == "go.mod" || filepath.Base(path) == "go.sum"
```
myitcv, line 144: I'm not sure this is entirely true in the context of CGO. Indeed the client can't really know what is relevant or not. And with `//go:embed` we could need to care about any file, because removal of a file that is the source of a `//go:embed` will result in an error.

So this would point to us having to notify about _everything_. Perhaps we address that in a follow up PR, along with batching the notifications?

myitcv (approved):
Some small nits, and one more fundamental questions. But overall, I'm very supportive and grateful for you tidying up this code. It's one of those areas with loads of edge cases. So better to get this in sooner rather than later, and then iterate of further fixes guided by tests.

cmd/govim/internal/fswatcher/os_others.go
leitzler: Duplicate comment :)

leitzler: Added an explanation.

cmd/govim/internal/fswatcher/os_others.go
leitzler: Fixed as per our Slack conversation & moved `populatedWatchers` into `fswatcher`.

cmd/govim/watcher.go
leitzler: Interesting, this is unchanged code but that's a good point I didn't think about. Let's do it with the batching.

cmd/govim/internal/fswatcher/os_darwin.go
leitzler: Good catch, it's a leftover from before when the go.mod path was passed in. We shouldn't watch the `root` parent, just the root. Fixed.

cmd/govim/internal/fswatcher/os_darwin.go
leitzler: Added a description for both versions of `New` and added a sanity check that ensure that the root is a directory.

myitcv:
Still LGTM 😄

PR #1058, opened 2021-03-15
cmd/govim: check quickfix list origin before update

tbruyelle:
QuickfixDiagnostics routine updates the quickfix list each time it
receives new diagnostics from gopls. This can be problematic if the
quickfix list has been populated by an other plugin, because the user
could be using this plugin.

The fix uses the quickfix title to determine if the quickfix is "owned"
by govim or by an other plugin. If the quicfix title is empty (which
means the quickfix has never been used before) or equal to "govim", then
govim considers it can change it, else it considers it cannot.

This condition is ignored when :GOVIMQuickfixDiagnostics is invoked by
the user, making possible for the user to display again the diagnostics
when he wants to.

### DIFF
--- a/cmd/govim/quickfix.go
+++ b/cmd/govim/quickfix.go
@@ -7,6 +7,10 @@ import (
 	"github.com/govim/govim"
 )
 
+const (
+	quickfixDiagnosticsTitle = "govim diagnostics"
+)
+
 type quickfixEntry struct {
 	Filename string `json:"filename"`
 	Lnum     int    `json:"lnum"`
@@ -24,32 +28,24 @@ func (q quickfixEntry) equalModuloBuffer(q2 quickfixEntry) bool {
 }
 
 func (v *vimstate) quickfixDiagnostics(flags govim.CommandFlags, args ...string) error {
-	wasNotDiagnostics := !v.quickfixIsDiagnostics
-	v.quickfixIsDiagnostics = true
-	return v.updateQuickfixWithDiagnostics(true, wasNotDiagnostics)
+	return v.updateQuickfixWithDiagnostics(true)
 }
 
 // updateQuickfixWithDiagnostics updates Vim's quickfix window with the current
-// diagnostics(), respecting config settings that are overridden by force.  The
-// rather clumsily named wasPrevNotDiagnostics can be set to true if it is
-// known to the caller that the quickfix window was previously (to this call)
-// being used for something other than diagnostics, e.g. references.
-func (v *vimstate) updateQuickfixWithDiagnostics(force bool, wasPrevNotDiagnostics bool) error {
-	if !force && (v.config.QuickfixAutoDiagnostics == nil || !*v.config.QuickfixAutoDiagnostics) {
+// diagnostics(), respecting config settings that are overridden by force.
+func (v *vimstate) updateQuickfixWithDiagnostics(force bool) error {
+	diags := v.diagnostics()
+	diagsHasChanged := v.lastDiagnosticsQuickfix != diags
+	canDiagnostics := v.quickfixCanDiagnostics()
+	autoDiag := v.config.QuickfixAutoDiagnostics == nil || *v.config.QuickfixAutoDiagnostics
+	v.lastDiagnosticsQuickfix = diags
+	if !force && (!diagsHasChanged || !canDiagnostics || !autoDiag) {
 		return nil
 	}
-	diagsRef := v.diagnostics()
-	work := v.lastDiagnosticsQuickfix != diagsRef
-	v.lastDiagnosticsQuickfix = diagsRef
-	if (!force && !work) || !v.quickfixIsDiagnostics {
-		return nil
-	}
-	diags := *diagsRef
 
 	// must be non-nil
 	fixes := []quickfixEntry{}
-
-	for _, d := range diags {
+	for _, d := range *diags {
 		// make fn relative for reporting purposes
 		fn, err := filepath.Rel(v.workingDirectory, d.Filename)
 		if err != nil {
@@ -73,15 +69,16 @@ func (v *vimstate) updateQuickfixWithDiagnostics(force bool, wasPrevNotDiagnosti
 	// by stashing the last selected diagnostic when we flip to, for example,
 	// references mode. But for now we keep it simple.
 	newIdx := 0
-	if !wasPrevNotDiagnostics && len(v.lastQuickFixDiagnostics) > 0 {
-		var want qflistWant
-		v.Parse(v.ChannelExpr(`getqflist({"idx":0})`), &want)
-		if want.Idx == 0 {
-			// NOTE: this should never happen since idx == 0 only if the qf is empty
-			// and we just tested it isn't.
+	if canDiagnostics && len(v.lastQuickFixDiagnostics) > 0 {
+		var qflist qflistProps
+		v.Parse(v.ChannelExpr(`getqflist({"idx":0})`), &qflist)
+		if qflist.Idx == 0 {
 			goto NewIndexSet
 		}
-		wantIdx := want.Idx - 1
+		wantIdx := qflist.Idx - 1
+		if len(v.lastQuickFixDiagnostics) <= wantIdx {
+			goto NewIndexSet
+		}
 		currFix := v.lastQuickFixDiagnostics[wantIdx]
 		var fileNextIdx, fileLastIdx, dirFirstIdx int
 		for i, f := range fixes {
@@ -116,17 +113,20 @@ func (v *vimstate) updateQuickfixWithDiagnostics(force bool, wasPrevNotDiagnosti
 		}
 	}
 NewIndexSet:
-	v.lastQuickFixDiagnostics = fixes
-	v.BatchStart()
-	v.BatchChannelCall("setqflist", fixes, "r")
-	if newIdx > 0 {
-		v.BatchChannelCall("setqflist", []quickfixEntry{}, "r", qflistWant{Idx: newIdx})
-	}
-	v.MustBatchEnd()
-
+	v.setQuickfixDiagnostics(fixes, newIdx)
 	return nil
 }
 
-type qflistWant struct {
-	Idx int `json:"idx"`
+// setQuickfixDiagnostics fills quickfix list with diagnostics, and set the title and index (if != 0).
+func (v *vimstate) setQuickfixDiagnostics(diags []quickfixEntry, index int) {
+	v.lastQuickFixDiagnostics = diags
+	v.BatchStart()
+	v.BatchChannelCall("setqflist", diags, "r")
+	v.BatchChannelCall("setqflist", []quickfixEntry{}, "r", qflistProps{Title: quickfixDiagnosticsTitle, Idx: index})
+	v.MustBatchEnd()
+}
+
+type qflistProps struct {
+	Idx   int    `json:"idx,omitempty"`
+	Title string `json:"title,omitempty"`
 }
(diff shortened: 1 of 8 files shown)


cmd/govim/quickfix.go
```diff
@@ -44,6 +44,12 @@ func (v *vimstate) updateQuickfixWithDiagnostics(force bool, wasPrevNotDiagnosti
 	if (!force && !work) || !v.quickfixIsDiagnostics {
 		return nil
 	}
+	var qflist qflistProps
+	v.Parse(v.ChannelExpr(`getqflist({"title":1})`), &qflist)
+	if !force && qflist.Title != "" && qflist.Title != "govim" {
```
tbruyelle, line 49: On thing I would like to improve is to merge this condition and the one from `v.quickfixIsDiagnostics`.

The later is set to false when Govim populates the quickfix with references, but we could use quickfix title also in that case.

I think if I turn `v.quickfixIsDiagnostics` into a function that returns true if the quickfix title is empty or equal to "govim" (maybe rename that to "govim-diagnostics", because we could add an other title for Govim references), then it would clean a little bit the code :

- only one condition here instead of 2
- no more manual `v.quickfixIsDiagnostics = true|false` since it relies on quickfix title

cmd/govim/testdata/scenario_default/quickfix_from_other.txt
```diff
@@ -0,0 +1,95 @@
+# Test that the quickfix window isn't updated by govim if it isn't populated by govim
+
+vim ex 'e main.go'
+vim ex 'call setqflist([{\"filename\":\"foo\"}, {\"filename\":\"bar\"}, {\"filename\":\"baz\"}], \"r\")'
```
myitcv, line 4: This is a great test.

One thing I think we need to do before calling this is verify that the quickfix list gets populated with the error diagnostics from `gopls`. That's not essential to this test, but instead a more realistic reflection of real life. So that way you end up with a call to `vimexprwait` before this call, then as you have it another one afterwards.

cmd/govim/testdata/scenario_default/quickfix_from_other.txt
```diff
@@ -0,0 +1,95 @@
...
+vimexprwait errors.other GOVIMTest_getqflist()
+# Duplicate error, the quickfix should remain because it hasn't been populated by govim
```
myitcv, line 6: If you put a blank line before this comment, [`testscript`](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript) will treat it as a section. In the verbose output from `testscript`, sections that complete "successfully" are collapse under the "heading" of the comment that precedes the contiguous lines of that section.

cmd/govim/testdata/scenario_default/quickfix_from_other.txt
```diff
@@ -0,0 +1,95 @@
...
+vim ex 'call cursor(6, 1)'
+vim ex 'call feedkeys(\"yyp\")'
+vimexprwait errors.other GOVIMTest_getqflist()
+# Force display diagnostics now
```
myitcv, line 10: Ditto.

cmd/govim/quickfix.go
myitcv: Yes I'd be happy to revisit this. Either in this PR or a later one. As you observed, it's rather clumsy now. Vim is essentially the source of truth for what is in the quickfix window at any given time, and your change helps move us closer to that arrangement.

myitcv (commented):
Generally LGTM with a couple of suggestions

cc @leitzler for thoughts too.

cmd/govim/testdata/scenario_default/quickfix_from_other.txt
tbruyelle: @myitcv OK, I added an assertion for the first error 35c475d

cmd/govim/testdata/scenario_default/quickfix_from_other.txt
tbruyelle: done 35c475d

cmd/govim/testdata/scenario_default/quickfix_from_other.txt
tbruyelle: done 35c475d

myitcv (approved):
LGTM - will leave it up to you @tbruyelle if you want to make the change you discussed in this PR or a follow up.

Thank you for your patience with this change!

tbruyelle:
@myitcv thx for the review. I'll try to see if I can implement the change we discuss simply, if yes I'll add a commit, if not I'll probably go with an other MR.

cmd/govim/quickfix.go
```diff
@@ -24,32 +28,29 @@ func (q quickfixEntry) equalModuloBuffer(q2 quickfixEntry) bool {
...
+func (v *vimstate) updateQuickfixWithDiagnostics(force bool) error {
 	if !force && (v.config.QuickfixAutoDiagnostics == nil || !*v.config.QuickfixAutoDiagnostics) {
 		return nil
 	}
-	diagsRef := v.diagnostics()
-	work := v.lastDiagnosticsQuickfix != diagsRef
-	v.lastDiagnosticsQuickfix = diagsRef
-	if (!force && !work) || !v.quickfixIsDiagnostics {
+
+	var (
```
myitcv, line 41: Please change this from a `var` block to short variable declarations.

myitcv (commented):
Thanks @tbruyelle - broadly speaking this looks fine, with the one requested changed for short variable declarations please. 

However, I'd like @leitzler to opine too, because this potentially influences our strategy with regards to a Vim grep workaround.

cmd/govim/quickfix.go
tbruyelle: @myitcv sure no problem, by curiosity why do you prefer this kind of declaration ? On my side I like to pack variable declarations like that, because that formats the code and I found it's more clear and readable.

cmd/govim/quickfix.go
```diff
@@ -24,32 +28,29 @@ func (q quickfixEntry) equalModuloBuffer(q2 quickfixEntry) bool {
...
+		diags                 = v.diagnostics()
+		diagsHasChanged       = v.lastDiagnosticsQuickfix != diags
+		quickfixIsDiagnostics = v.quickfixIsDiagnostics()
+	)
+	v.lastDiagnosticsQuickfix = diags
+	if (!force && !diagsHasChanged) || (!force && !quickfixIsDiagnostics) {
```
leitzler, line 47: Really nits, but I personally think:
```go
if !force && (!diagsHasChanged || !quickfixIsDiagnostics) {
```
would be slightly more readable and consistent with other force checks in the code.

leitzler (approved):
First of all, nice feature!

When it comes to vimgrep this slightly alters the behaviour (apart from the obvious, that the existing matches doesn't disappear when diagnostics arrive). The behaviour that do change is if the quickfix list is "owned" by someone but empty. 

e.g.
1) There are no govim diagnostics
2) User calls vimgrep but finds no matches, it returns with a `E480: No match: foobar` message
3) The user introduce a new error, but the quickfix remains empty since the title changed, even if it wasn't populated

In my workflow I use `:cw` to open the quickfix list, and it might be confusing if it "doesn't work" as it isn't clear that the title changed at step 2. I think it make sense if `govim` actually did hi-jack the quickfix if the list is empty.

Apart from that and a tiny nit it LGTM.

cmd/govim/quickfix.go
tbruyelle: @leitzler ofc ! I need to revise my boolean algebra courses :)

d3c8ef1

tbruyelle:
> In my workflow I use `:cw` to open the quickfix list, and it might be confusing if it "doesn't work" as it isn't clear that the title changed at step 2. I think it make sense if `govim` actually did hi-jack the quickfix if the list is empty.

@leitzler What if the new `isQuickfixDiagnostics()` function returns true if the quickfix list is empty ?

Maybe we should rename that function seems it's not really `isQuickfixDiagnostics` but more likely `canQuickfixDiagnostics`... WDYT?

cmd/govim/quickfix.go
myitcv: I've honestly not seen much (any?) use of parenthesised `var` declaration lists "outside" of top level file declarations. At the file level they do have things to stand out more, and when combined with `import` and `const` declarations, visually make sense. But within a function, they're a bit clumsy and unnecessarily verbose to my mind: which is precisely what short variable declarations exist to help with (amongst other situations like `for` loops).

tbruyelle:
@leitzler @myitcv While I was trying some other simplifications in `updateQuickfixWithDiagnostics`,  I've spotted a bug in this change :

If I remove those lines : 
```diff
 func (v *vimstate) updateQuickfixWithDiagnostics(force bool) error {
-       if !force && (v.config.QuickfixAutoDiagnostics == nil || !*v.config.QuickfixAutoDiagnostics) {
-               return nil
-       }
```

The test ` TestScripts/scenario_default/quickfix_config` is still green, while it should be red, because `quickfixAutoDiagnostics` is ignored.

To make it red as it should be, I have to update the code in `vimstate.go`, which handles the config change of this particular field. Indeed, the code was updating the quickfix list with empty diagnostics when `autoDiagnostics` was set to false, but without setting the quickfix title. If it set the title as expected, the test goes red.

So the fix introduces a new function `setQuickfixDiagnostics` that populates the quickfix with diagnostics passed in parameters and sets the expected title.

See c977523

cmd/govim/quickfix.go
tbruyelle: Ok, as you wish b65554c

myitcv:
@tbruyelle thanks for making the various changes. I'll hand over to @leitzler at this point for you both to round out the discussion above.

leitzler:
> > In my workflow I use `:cw` to open the quickfix list, and it might be confusing if it "doesn't work" as it isn't clear that the title changed at step 2. I think it make sense if `govim` actually did hi-jack the quickfix if the list is empty.
> 
> @leitzler What if the new `quickfixIsDiagnostics()` function returns true if the quickfix list is empty ?

Yes, sounds good.

> Maybe we should rename that function seems it's not really `quickfixIsDiagnostics` but more likely `quickfixCanDiagnostics`... WDYT?

I agree, that would describe the function better.

tbruyelle:
> > > In my workflow I use `:cw` to open the quickfix list, and it might be confusing if it "doesn't work" as it isn't clear that the title changed at step 2. I think it make sense if `govim` actually did hi-jack the quickfix if the list is empty.
> > 
> > 
> > @leitzler What if the new `quickfixIsDiagnostics()` function returns true if the quickfix list is empty ?
> 
> Yes, sounds good.
> 
> > Maybe we should rename that function seems it's not really `quickfixIsDiagnostics` but more likely `quickfixCanDiagnostics`... WDYT?
> 
> I agree, that would describe the function better.

@leitzler I have updated the function accordingly, and added a test. See 4edf3be

Now if the quickfix is empty, it can be updated by diagnostics, whatever its title.

leitzler:
Looks good to me!

Please squash the commits as well before merge (as per https://github.com/govim/govim/wiki/Contributors#pull-request-contributions) .

Thank you very much for all the updates (and the patience)!

tbruyelle:
> Looks good to me!
> 
> Please squash the commits as well before merge (as per https://github.com/govim/govim/wiki/Contributors#pull-request-contributions) .

done 2ed54c7 

> Thank you very much for all the updates (and the patience)!

No problem, it was a pleasure to contribute to this great project :)

PR #1066, opened 2021-03-27
general: fix gitattributes

myitcv:
Per https://github.com/github/linguist/issues/5288 we were not using the
recursive form of a pattern to specify that everything under
cmd/govim/internal/golang_org_x_tools should be considered generated.


myitcv:
Tested with my first push to this branch.

PR #1068, opened 2021-03-27
cmd/govim: do not watch file system when there is no module root

### LINKED ISSUE #1067: cmd/govim: module detection do not detect "module mode w/o module root"

<pre>
$ go version
go version devel +d8e33d558e Fri Feb 26 18:57:20 2021 +0000 linux/amd64
</pre>

$ go env
N/A
</pre></details>

Current main, a68991d

We currently detect if `govim` is running in "module mode" before creating a file system watcher:
https://github.com/govim/govim/blob/a68991df886ed6969eae788e5f447833ebfdfe1d/cmd/govim/gopls.go#L186-L187

But as per [CL 148517](https://go-review.googlesource.com/c/148517) `go env GOMOD` returns `os.DevNull` if we are in module mode without a module root.

This is the case in some of our tests, one of them being `TestScripts/scenario_default/buffers.txt`. When those tests run we create a file system watcher that watches anything that happens in `/dev/` since we think that `/dev/null` is the `go.mod` file.

It could lead to **a lot** of events when, for example, using `ssh` into a remote machine it continuously writes to `/dev/pts/...` as things are written to the terminal.

We should skip creating a file watcher if the detected `go.mod` file is `os.DevNull`.

leitzler:
If we are using module mode but isn't inside a module "go env GOMOD"
will return os.DevNull (e.g. /dev/null in linux).

That did incorrectly add a file watcher to /dev since govim thought
/dev/null was the go.mod file.

Now we skip creating a fs watcher if we can't find the module root.

Fixes #1067

### DIFF
--- a/cmd/govim/gopls.go
+++ b/cmd/govim/gopls.go
@@ -183,7 +183,9 @@ func (g *govimplugin) startGopls() error {
 		return fmt.Errorf("failed to derive go.mod path: %v", err)
 	}
 
-	if gomodpath != "" {
+	// "go env GOMOD" might return an empty string (not module mode) or os.DevNull (in module
+	// mode but without a module root).
+	if gomodpath != "" && gomodpath != os.DevNull {
 		// i.e. we are in a module
 		mw, err := newModWatcher(g, gomodpath)
 		if err != nil {


myitcv (approved):
Good catch - LGTM

PR #1069, opened 2021-03-29
cmd/govim: extend GOVIM_LOG to accept multiple commaseparated values

leitzler:
This change adds the possibility to set multiple values when using
`GOVIM_LOG`. It is mainly a preparation for the upcoming file watcher
refactoring where the intention is to add opt-in for more verbose
logging.

### DIFF
--- a/cmd/govim/main.go
+++ b/cmd/govim/main.go
@@ -91,7 +91,7 @@ func launch(goplspath string, in io.ReadCloser, out io.WriteCloser) error {
 	var logFile *os.File
 	var writers []io.Writer
 
-	if getEnvVal(d.goplsEnv, config.EnvLog, "on") == "on" {
+	if d.logging["on"] {
 		logFile, err = d.createLogFile("govim")
 		if err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


cmd/govim/main.go
```diff
@@ -91,7 +91,8 @@ func launch(goplspath string, in io.ReadCloser, out io.WriteCloser) error {
 	var logFile *os.File
 	var writers []io.Writer
 
-	if getEnvVal(d.goplsEnv, config.EnvLog, "on") == "on" {
+	logging := csvToMap(getEnvVal(d.goplsEnv, config.EnvLog, "on"))
```
myitcv, line 94: We should build this map once and set it on the `govimplugin`. Which would also accurate reflect the semantics that we do not mean to support the semantics of the values changing.

myitcv (approved):
LGTM - feels like we're getting close to needing tests for this?

myitcv:
> Any ideas on how to test it in a good way? 

We should be able to use testscript tests, and set `GOVIM_LOGFILE_TMPL=%v` in order that we have deterministic file names for log files. The tricky (but very much doable) part will be that we need to setup environment variables as part of the `Setup` phase of a testscript test. That will probably require that we extend our concept of per-scenario `user_config.json` files to also include per-scenario `env.json` files, which are simply a `map[string]*string` of environment values. These values would be "set" as part of the `Setup` phase, and also passed in the call to `newplugin` (the parameter is currently called `goplsEnv`, but it need not be restricted/exclusive to that)

> Might defer the testing to a later PR if you don't mind?

That's fine, just wanted to kick off a discussion for us to think about it.

PR #1075, opened 2021-04-09
cmd/govim: do not wait more than 500ms for gopls Shutdown

myitcv:
As explained in golang.org/issue/45476, all call to Shutdown does not
properly interrupt "work" that might be in progress in gopls. This has
the effect of blocking an exit from Vim.

Workaround this for now by limiting the wait time to 500ms, which is
effectively the measure of a user's patience before anger/frustration
sets in!

### DIFF
--- a/cmd/govim/main.go
+++ b/cmd/govim/main.go
@@ -3,6 +3,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


myitcv:
cc @mvdan

mvdan:
Sounds good to me. I'd even find 1s acceptable, but anything beyond that definitely not :)

leitzler (approved):
LGTM

PR #1078, opened 2021-05-06
deps: update x/tools & gopls to c0140e85

leitzler:
This change also includes moving ShowDocument from server to client
according to 062bf4eb (CL 315990).

* go/analysis/passes/printf: adjust error message c0140e85
* internal/lsp/source: support typeDefinition for function/method's return values d1ea2c78
* internal/lsp: handle exclude directives in multi-module mode 08a4f343
* internal/lsp/regtest: add a benchmark for didChange f4a41298
* lsp/source: enable experimentalPostfixCompletions by default dd255f23
* passes/stdmethods: warn when an Is, As, or Unwrap has the wrong signature. 19496731
* lsp/completion: improve append() param type inference 68c6cab8
* internal/lsp: add semantic tokens for comments and multiline strings 250398d0
* internal/lsp: support template files 7cab0ef2
* gopls/go.mod: upgrade to Go 1.17 f03daeac
* internal/lsp/regtest: run one quick fix at a time in TestUnknownRevision 42984c42
* go.mod: upgrade to Go 1.17 19b1717e
* all: upgrade go.mod files to Go 1.16 a1fbb681
* internal/lsp: make ShowDocument RPC available to gopls 062bf4eb
* internal/lsp: warn users who have built gopls with go-diff v1.2.0 3e17c62e
* gopls: clarify policy with respect to support for older Go versions def02638
* internal/lsp: don't use ast.NewPackage to build builtin 7a6108e9
* internal/lsp/completion: indicate completion candidates that are deprecated edbe9bef
* internal/lsp/regtest: force GOPACKAGESDRIVER=off 9b9633e0
* gopls/internal/regtest/modfile: set an explicit go version in the TestUnknownRevision modules 291330a8

### DIFF
--- a/cmd/govim/gopls_client.go
+++ b/cmd/govim/gopls_client.go
@@ -40,6 +40,11 @@ const (
 
 var _ protocol.Client = (*govimplugin)(nil)
 
+func (g *govimplugin) ShowDocument(context.Context, *protocol.ShowDocumentParams) (*protocol.ShowDocumentResult, error) {
+	defer absorbShutdownErr()
+	panic("ShowDocument not implemented yet")
+}
+
 func (g *govimplugin) ShowMessage(ctxt context.Context, params *protocol.ShowMessageParams) error {
 	defer absorbShutdownErr()
 	g.logGoplsClientf("ShowMessage callback: %v", pretty.Sprint(params))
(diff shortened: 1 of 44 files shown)


myitcv (approved):
LGTM

PR #1081, opened 2021-05-24
deps: update x/tools & gopls to 40613125

myitcv:
Includes some test assertion fix ups following
https://go-review.googlesource.com/c/tools/+/301949

* internal/lsp: add list_known_packages and add_import commands 40613125
* tools: make printf analysis have more helpful output 1e0c9609
* internal/lsp: correct typo in orphaned file error message 2275bb55
* internal/lsp/semantic: only return token types the client asked for 30637901
* internal/span: compute utf16 length directly 0886cdd1
* internal/lsp/cache: don't delete metadata until it's reloaded 46e69bf3
* internal/lsp/source: refactor commentToMarkdown function f803486e


leitzler (approved):
LGTM

Issue #1083, opened 2021-05-28
cmd/govim: properly handle and support RegisterCapability for workspace/didChangeConfiguration method


myitcv:
cc @leitzler

PR #1084, opened 2021-05-28
deps: update x/tools & gopls to 5ab822f6

myitcv:
* internal/lsp: enable semantic tokens by default 5ab822f6
* internal/lsp: findIdentifier always return an error eb0b8a15


myitcv:
@leitzler please note the workaround I've added as part this, linked to https://github.com/govim/govim/issues/1083

myitcv:
> if we decide to include it)

You mean "when we include it" ;-)

PR #1085, opened 2021-05-30
cmd/govim: update reference to gopls analyzer documentation in config.go

leitzler:
The link to gopls documentation describing analyzers was broken. It now
points to the correct page.


myitcv (approved):
X-cellent.

PR #1088, opened 2021-06-07
cmd/govim: use span.URI.Filename() when handling filenames

myitcv:
When interpretting URIs from gopls, we need to use the
span.URI.Filename() when we are expecting a filename. This ensures we
properly decode the URL, which in turn properly decodes the hex-encoded
path elements, like '!', which can appear in module paths.


leitzler (approved):
LGTM

PR #1090, opened 2021-06-21
cmd/govim: adapt to changes in Vim v8.2.3019

myitcv:
In Vim v8.2.3019, locations now include end position information,
documented as follows:

   end_lnum - end of line number if the item is multiline
   end_col - end of column number if the item has range

Update all of our golden test output to reflect this change. All values
are 0 because we do not have any range-based values.


leitzler (approved):
LGTM

PR #1093, opened 2021-07-15
cmd/govim: add support for semantic token highlighting

leitzler:
In LSP there is a concept of "semantic tokens" that gopls can provide
us. It allows us to do a more granular syntax highlighting than vim
provides default for Go.

This is a first take of how it might work, used to try out to what
extent it makes sense to govim to provide syntax highlighting.

The highlight groups are currently defined as a part of the syntax
highlighting so to try it out you currently have to have syntax
enabled/on when loading vim. Semantic tokens will replace the native
syntax. Turn off syntax (":syntax off") after startup to see semantic
tokens only.

Enable by adding to your .vimrc:

```
call govim#config#Set("ExperimentalSemanticTokens", 1)
```

This change aims to keep track of all visible lines in vim, and only
request semantic tokens for those ranges. Each time one of the
autocommands CursorHold, CursorHoldI or CursorMoved fires visible
lines are re-evaluated. Since highlighting is very sensitive to
latency (and CursorMoved triggers A LOT) the delta calculation is
done on the vim side. That way govim only sees changes.

In govim we also debounce multiple updates, and keep track of
which lines we placed tokens on, to be able to remove non-visible
tokens. Since a buffer might be visible in several windows we also
ensure that we do not remove tokens visible in other windows.


leitzler:
Example of how it looks like:

[asciicast](https://asciinema.org/a/qptU4OF6u04tbgRMFB9fInxqC)

leitzler:
[asciicast](https://asciinema.org/a/fJwiid8RzAVzAzBXDxsHsWc9v)

Demo with a larger file (~100k, the limit in gopls).

PR #1096, opened 2021-07-30
deps: upgrade LatestVim to be v8.2.3249

myitcv:


_scripts/publishDockerImages.sh
```diff
@@ -2,4 +2,4 @@
 
 source "${BASH_SOURCE%/*}/common.bash"
 
-docker push govim/govim
+docker push -a govim/govim
```
leitzler, line 5: What's the rational behind this change?

testdriver/testdriver.go
```diff
@@ -887,6 +887,17 @@ func Condition(cond string) (bool, error) {
 	if !semver.IsValid(v) {
 		return false, fmt.Errorf("%v is not a valid semver version", v)
 	}
+	// We have version information. Be perhaps a bit overly
```
leitzler, line 890: Nice change. Maybe we can phrase it a bit different to differentiate `vim` and `vim:v8.2.x`? I think that a future me might be confused when reading tests.

I don't have a nice suggestion on how to improve though. `vimver:v8.2.xxxx` for version?
Or `flavor:vim` / `flavor:gvim`?

Something that differentiate version `vim` from flavour `vim` would be nice.

_scripts/publishDockerImages.sh
myitcv: Apparently a change in behaviour of `cmd/docker`... at least it appears that way to me. I didn't dig too far, just fixed things up :)

testdriver/testdriver.go
myitcv: Thanks. It wasn't sitting quite right with me either. So I simplified things even further in https://github.com/govim/govim/pull/1098. Please let me know what you think. The tests hopefully make clear what's going on now too.

leitzler (approved):
LGTM

PR #1099, opened 2021-08-03
deps: update x/tools and gopls to 3395cb03

myitcv:
CL 338789 fixed a bug on the gopls side with respect to finding
references. This highlighted a couple of bad tests on the govim side: we
where trying to jump to definition with the cursor over the '.' in, say,
'time.Kitchen' - this is ambiguous at best.

Update to the latest tools/gopls and fix our test assertions in the
process, by adjusting the cursor position to the right by one.

* go/types/objectpath: minor doc fix 3395cb03
* internal/lsp: send "extract variable" edits ordered bb69444e
* internal/lsp: handle invalid positions in semantic token debug logic f8cfadac
* internal/lsp: remove duplicated loop that copies IDs 45eff0fd
* internal/lsp: find references for ident before selector a6684984
* go/ssa/interp: handle nil slice convert to array pointer 3810fa82
* go/pointer: fix constraint gen for *ssa.Next ab1fe720


leitzler (approved):
LGTM

PR #1101, opened 2021-08-18
deps: update x/tools and gopls to bf6c7f26e9f0

leitzler:
* all: update dependencies after gopls/v0.7.1 release bf6c7f26
* go/analysis/passes/printf: fix %w for non-fmt.Errorf functions e32504ce
* internal/lsp/cache: fix flakiness of load progress reports 96f147c2
* internal/lsp/source: recursively search for qualified objects 6932d22a
* x/tools/analysis/passes/printf: fix error message for unsuppported %w 10bcabde
* internal/typeparams: fix undefined *types.Named.TArgs method 758a1a1d
* internal/lsp: export and move some objects related to metadata a55d5155
* internal/lsp: add support for go.work files in file watching 45389f59
* go/analysis/passes/testinggoroutine: also inspect defined/anonymous functions 8fae06a8
* go/analysis: add slice to array pointer conversion to nilness 3fce476f
* go/pointer: support ssa.SliceToArrayPointer 03a91dd9
* internal/lsp/source/completion: exclude 'any' from lexical results d52cb71c
* internal/lsp: allow for multiple ad-hoc packages in the workspace 9f495611
* lsp/completion: reduce garbage searching for candidates 15a2ab85
* gopls/internal/regtest: fix completion benchmarks 749a5cd4
* internal/lsp: use LookupParent for finding scope f91c4a58
* internal/lsp/semantic.go: continue even without type information 5f5a173a
* internal/lsp: improve error message about unsaved files 92b2fbe7


testsetup/testsetup.go
```diff
@@ -155,3 +156,26 @@ func EnvLookupFlavorCommand() (flav govim.Flavor, cmd Command, err error) {
...
+// the first vim found is the subcommand.
+func RealVimPath() (path string, err error) {
+	var firstFound bool
+	for _, dir := range filepath.SplitList(os.Getenv("PATH")) {
+		path := filepath.Join(dir, "vim")
+		d, err := os.Stat(path)
+		if err != nil {
+			continue
+		}
+		if m := d.Mode(); m.IsDir() || m&0111 == 0 {
```
myitcv, line 172: Strictly speaking we only need the execute bit set.

myitcv (approved):
LGTM with one small suggestion.

testsetup/testsetup.go
leitzler: True. I hope you don't mind me leaving it as is, let me know if so and I'll push another PR.

The rationale is that directories are created with execution bit set by default, so if we (sometime in the future) have a directory named "vim" in the path it will cause a subtle bug that could be tricky to find.

PR #1103, opened 2021-09-09
plugin/govim: drop .log suffix on vim channel log

leitzler:
Some flavors of "mktemp", at least macOS and possibly OpenBSD) require
that the template ends with X's (instead of just having them anywhere in
the template). On macOS they won't be converted to a random number at
all.

This change drops the .log extension from the vim channel log. We could
have run mktemp in "dry-run" mode to just generate a file name, but that
actually creates and removes the file on macOS. That introduces a race
that makes it fairly easy for other users on the same system to "hijack"
the logfile.

### DIFF
--- a/plugin/govim.vim
+++ b/plugin/govim.vim
@@ -31,7 +31,6 @@ if index(split(system("echo -n ${GOVIM_LOG-on}"), ","), "on") >= 0
   if s:filetmpl == ""
     let s:filetmpl = "%v_%v_%v"
   endif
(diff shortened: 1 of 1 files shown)


myitcv (approved):
LGTM - on the assumption that someone somewhere has verified this?! 😄

leitzler:
Indeed, I verified that it works on macOS. 0xjmml also confirmed that it works on OpenBSD over at Gopher's Slack in #govim :)

PR #1109, opened 2021-09-22
cmd/govim: support multi-lined diagnostics in hover

leitzler:
Type parameter support in gotip has a few diagnostics that is
multi-lined. We now properly show them in hover popups as multiple
lines, and appends the source to the first line only.


myitcv (commented):
LGTM assuming you've tested this locally?

Ideally would be nice to get some sort of automated coverage for things like this if possible?

myitcv (approved):

leitzler:
Yes, tested locally. Tried to find any multi-lined diagnostics pre-Go1.18 to add as test but only found type parameter related ones :(

leitzler:
Wrote #1111 as a reminder to myself.

PR #1124, opened 2021-11-29
cmd/govim: Support setting gopls memory mode

leitzler:
In gopls there is an experimental feature that can be used to reduce the
memory footprint by degrade features in packages without any open files.
This can be useful in large single-moduled mono repos that otherwise
consume lots of resources.

We support the both modes defined in gopls, "Normal" and "DegradeClosed"
where "Normal" is the default behaviour. To use "DegradeClosed" add a
line to your .vimrc:

```
call govim#config#Set("ExperimentalGoplsMemoryMode","DegradeClosed")
```

### DIFF
--- a/autoload/govim/config.vim
+++ b/autoload/govim/config.vim
@@ -206,6 +206,14 @@ function! s:validExperimentalWorkspaceModule(v)
   return s:validBool(a:v)
 endfunction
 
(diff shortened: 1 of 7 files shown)


myitcv (approved):
LGTM - thanks for your patience.

PR #1129, opened 2022-01-04
_scripts: build docker images using --secret

myitcv:
This ensures that we don't expose build args as part of the resulting
image.

### DIFF
--- a/_scripts/Dockerfile.gvim
+++ b/_scripts/Dockerfile.gvim
@@ -6,10 +6,11 @@ RUN apt-get update && \
 
 RUN git config --global advice.detachedHead false
 
-ARG GH_USER
-ARG GH_TOKEN
-RUN echo -e "machine github.com\n  login $GH_USER\n  password $GH_TOKEN" >> ~/.netrc
-RUN echo -e "machine githubusercontent.com\n  login $GH_USER\n  password $GH_TOKEN" >> ~/.netrc
+RUN --mount=type=secret,id=GH_USER --mount=type=secret,id=GH_TOKEN \
+  export GH_USER=$(cat /run/secrets/GH_USER) && \
+  export GH_TOKEN=$(cat /run/secrets/GH_TOKEN) && \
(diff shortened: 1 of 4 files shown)


_scripts/Dockerfile.gvim
```diff
@@ -24,7 +25,8 @@ ENV PATH=/vim/bin:$PATH
 RUN rm ~/.netrc
 
 ARG GO_VERSION
-RUN curl -sL https://dl.google.com/go/${GO_VERSION}.linux-amd64.tar.gz | tar -C / -zx
+ARG TARGETARCH TARGETOS
```
leitzler, line 28: Leftovers from multi-arch experiments? Can't see that they are set anywhere.

_scripts/Dockerfile.vim
```diff
@@ -24,7 +25,8 @@ ENV PATH=/vim/bin:$PATH
 RUN rm ~/.netrc
 
 ARG GO_VERSION
-RUN curl -sL https://dl.google.com/go/${GO_VERSION}.linux-amd64.tar.gz | tar -C / -zx
+ARG TARGETARCH TARGETOS
```
leitzler, line 28: Same

leitzler (approved):
Just a comment, otherwise LGTM

_scripts/Dockerfile.gvim
myitcv: Exactly.

myitcv: Actually, just to check. Are you referring to `TARGET(OS|ARCH)` here?

If so, they are used on the following line.

Or did I misunderstand?

leitzler: Yes, as per our Slack conversation I didn't know that they were set automatically by buildkit. All good!

myitcv: Thanks for clarifying!

PR #1131, opened 2022-01-09
cmd/govim: handle multi-op fsevents

leitzler:
Events from fsnotify might hold several "file operations" per event as
operation is a bit mask. It hasn't been observed that Linux ever do, but
it has been seen when running Darwin. Since govim doesn't use fsnotify
in darwin (we use fsevents), we haven't seen this fail in our tests.

This change checks operation bit masks so that combined operations
aren't lost.


leitzler:
As kindly pointed out by @rogpeppe, thanks a lot!

myitcv (approved):
LGTM

PR #1136, opened 2022-02-07
cmd/govim: disable allowModfileModifications by default

### LINKED ISSUE #1112: turn ExperimentalAllowModfileModifications off by default
https://github.com/golang/tools/blob/master/gopls/doc/settings.md says:

```
allowModfileModifications bool

This setting is experimental and may be deleted.

allowModfileModifications disables -mod=readonly, allowing imports from out-of-scope modules. This option will eventually be removed.

Default: false.
```

govim seems to have it on by default:

https://github.com/govim/govim/blob/f1143596a249a83b41afc5209c4320b8fc7b217b/cmd/govim/main.go#L254

The fact that gopls discourages its use should be enough reason to not turn it on by default for all users. Plus, it seems to cause confusing errors at times, such as https://github.com/golang/go/issues/48186.

leitzler:
The previous default value for for the config setting
ExperimentalAllowModfileModifications was "true", which isn't the
default value for gopls.

Enabling this setting causes issues with the new workspace mode in
gopls as described in #1112 and referred golang/go#48186. It also seem
to cause issues in upcoming Go 1.18, see golang/go#51056.

Closes #1112

### DIFF
--- a/cmd/govim/testdata/scenario_default/go_to_def_upper_case.txt
+++ b/cmd/govim/testdata/scenario_default/go_to_def_upper_case.txt
@@ -28,7 +28,10 @@ module mod.com
 
 go 1.12
 
-require example.com/blah v1.0.0
+require example.com/blaH v1.0.0
+-- go.sum --
+example.com/blaH v1.0.0 h1:iXP7NTq9zM+XteJ+LOgQ4/86Qk9gcwGy7izwrf03Hng=
+example.com/blaH v1.0.0/go.mod h1:2yAkoKJQQy5q8VEphynudWTTk7rzAUrdv+/+RxewcXI=
 -- p.go --
 package p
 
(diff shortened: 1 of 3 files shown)


cmd/govim/testdata/scenario_default/go_to_def_upper_case.txt
```diff
@@ -28,7 +28,10 @@ module mod.com
 
 go 1.12
 
-require example.com/blah v1.0.0
+require example.com/blaH v1.0.0
```
leitzler, line 31: I think this change is a bug fix, I'm I right? It just accidently worked before because `go.mod` got updated automatically to include both `blah` and `blaH` when updates to `go.mod` was allowed.

leitzler:
Commit reviewed & approved as part of #1137

PR #1140, opened 2022-02-24
plugin: reinstall when Go version changes

### LINKED ISSUE #1135: reinstall gopls (and govim?) binaries if the Go version changes
govim 1da33d23886a18dd8280770dad0d3ae60e872e6a

    go version devel go1.18-d588f48770 Sat Feb 5 06:14:58 2022 +0000 linux/amd64

    VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jan 16 2022 12:09:56)
    Included patches: 1-4106

I switch back and forth between `go` in my PATH being master (version above currently), and the latest stable (like 1.17.6). This normally works just fine for tools and builds.

And it does normally work for govim too, except for one edge case: if I start up govim from scratch with Go 1.17.6, it will build gopls with that version. If I then switch to Go master to edit some generic code, gopls will not work properly, as it doesn't know how to handle the generics syntax. And govim won't automatically rebuild gopls with my new current Go version, which does support generics.

@leitzler points out I could set `GOVIM_ALWAYS_INSTALL=true` to force govim to always `go install` gopls, but I reckon this should work by default. I seem to understand that, right now, govim will only rebuild gopls if the gopls commit hash in use changes. The simplest fix I can think of is to also rebuild gopls if the output of `go version` changes.

leitzler:
We now consider Go version in addition to govim commit when we determine
if govim and gopls should be reinstalled.

This helps when jumping between Go versions or following along with
an often updated gotip.

Fixes #1135

### DIFF
--- a/plugin/govim.vim
+++ b/plugin/govim.vim
@@ -409,7 +409,12 @@ function s:install(force)
     throw commit
   endif
 
-  let targetdir = s:plugindir."/cmd/govim/.bin/".commit."/"
+  let goversion = trim(system("go version"))
+  if v:shell_error
+    throw goversion
+  endif
+
+  let targetdir = s:plugindir."/cmd/govim/.bin/".sha256(commit.goversion)."/"
   if $GOVIM_DEBUG_INSTALL == "true"
 	  " The first line may directly follow a "hint" message from Vim.
 	  echo "\n"


mvdan (approved):

mvdan:
I rebuild tip every other day or so, so this will be massively useful :)

PR #1149, opened 2022-03-30
plugin: disable GOWORK during installation of gopls

leitzler:
The automatic installation of gopls used any present GOWORK which
made the installation fail. We now explicitly disable it.

### DIFF
--- a/plugin/govim.vim
+++ b/plugin/govim.vim
@@ -430,7 +430,7 @@ function s:install(force)
     echom "Installing govim and gopls"
     call feedkeys(" ") " to prevent press ENTER to continue
     " TODO: make work on Windows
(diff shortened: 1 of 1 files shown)


mvdan (approved):
While you're here, I imagine you can now remove GO111MODULE=on, as that's been the default since Go 1.16. The x/tools repository, including gopls, only supports two major Go versions, so they already require Go 1.17 or later.

leitzler:
Good point! I'm going to leave it in for now since we mention in the README that we support Go 1.12. That probably isn't true, and the GitHub actions doesn't test that old versions. I'll raise a separate issue for evaluating & updating minimal supported version. Thanks for the review!

PR #1151, opened 2022-04-05
cmd/govim: add signaturehelp support for type parameters

leitzler:
Go 1.18 introduced two new ast nodes that our signature help didn't
recognize, which led to panics. Those have added along with a new test.


mvdan (approved):
s/have/are/ in the commit message

Also, perhaps https://pkg.go.dev/golang.org/x/exp/typeparams would let you avoid the build tags.

leitzler:
Oh, didn't think about that package. It seems like it is missing an alias for `ast.IndexExpr` so I don't think we can avoid the build tags anyway(?)

leitzler:
Way cleaner, thanks!

Issue #1130, opened 2022-01-05
Support multi-arch docker images


myitcv:
Up until now, all development has happened on `linux/amd64` platform machines. With the arrival of M1 chip Apple machines, this has changed and development regularly happens on `linux/arm64` platform machines.

We should therefore support Docker images for those platforms. 

This will require:

* Upgrading base Docker images to Ubuntu 20.04 (required because `linux/arm64` docker images do not exist for Ubuntu 18.04)
* Working out why the `gvim` tests fail on Ubuntu 20.04 (initial investigation into this dropped)

This should like happen under `v2` of our images so as not to break existing builds. 

cc @leitzler

PR #1152, opened 2022-04-06
ci: update go and vim versions

leitzler:
Update go versions to ["go1.17.8", "go1.18"].
Update latest vim version to v8.2.4698.


.github/workflows/docker-based_tests.yml
```diff
@@ -35,9 +35,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-18.04]
-        go_version: ["go1.16.13", "go1.17.6", "go1.18beta1"]
+        go_version: ["go1.17.8", "go1.18"]
```
mvdan, line 38: why not use v1.17.x and v1.18.x, by the way? I see no reason to control the bugfix updates with precision, as they should never break any API or CLI command.

mvdan (approved):
FYI the same could be said of the OS version, e.g. ubuntu-latest, though that *does* imply big jumps in versions that could potentially break something like an `apt install` command.

.github/workflows/docker-based_tests.yml
leitzler: We push version specific docker images to docker hub (as a manual step) and I'm not sure how well that would play with an "x-version" (see https://hub.docker.com/r/govim/govim/tags).

Regarding OS version, Paul had some comments in govim/govim#1130 where he had tested newer ubuntu and some tests failed.

I'll leave it as is and wrote govim/govim#1153 covering wildcard versions.

mvdan: SGTM! replied on the new issue.

PR #1177, opened 2023-02-05
cmd/govim: remove ExperimentalWorkspaceModule support

leitzler:
The experimental support for workspace module has been removed from gopls (as per golang/go#55331). This change removes the setting from govim as preparation before updating gopls to a version that doesn't support workspace module.


myitcv (approved):
LGTM

PR #1178, opened 2023-02-09
cmd/govim: remove ExperimentalGoplsMemoryMode support

leitzler:
The experimental support for setting memory mode in gopls will soon be deprecated and then removed. Current test for it in govim wasn't foolproof and relied on the internals of gopls not changing. In the latest version of gopls it has already making the test fail.

Instead of coming up with an idea on how to test it now, and implementing that, we simply remove the feature ahead of time.


myitcv (approved):
LGTM

PR #1184, opened 2023-05-09
deps: update x/tools and gopls to 90e9c65995cc

myitcv:
As part of this, update an error message assertion that has changed as a result of the printf go/analysis checker changing in x/tools.


myitcv:
Superseded by #1185

PR #1207, opened 2024-06-12
deps: update x/tools and gopls to f45db9b4bbd0

myitcv:
(see commit message)


myitcv:
@leitzler - note I've skipped codelens_gc_details.txt in this PR because the test is now failing and I can't immediately see why. Any chance you could help take a look?

myitcv:
Might actually be a bug in `gopls`: https://gophers.slack.com/archives/CRWSN9NCD/p1718203404457789

myitcv:
Actually, it was a `govim` "bug". Fix included in this PR.

myitcv:
This sporadic failure of https://github.com/govim/govim/blob/86751534e9865627da41cf1178124759c820b31e/cmd/govim/testdata/scenario_default/quickfix_new_files.txt#L32 is suspicious. Leaving open until we know more.

myitcv:
Discussed with @leitzler and @findleyr via Slack. We're still none the wiser on the cause of this 50% test failure rate. Will require some more close debugging, capturing gopls log files for the failing test. At least I can semi reproduce this locally using `stress` to simulate load on my machine. But for now we're going to merge this PR to see if actual usage of the new `gopls` gives us better clues as to what the problem might be.

PR #1219, opened 2025-08-19
all: patch to work with Go 1.25

myitcv:
(see commit message)


tbruyelle:
:cry: for the maintenance mode...
@myitcv Did you switch to something else for your go LSP client ?

